PR #26, opened 2022-12-05
RFC - proof of concept of using a single error type

doom-weaver:
## Problem

We discussed offline having the javascript driver have a single error type instead of exposing multiple types.

## Solution

Implement that option.

## Result

Give some insight into the usability of a single error model vs a multi-error model.

## Testing

Ran full test suite - all pass.

## My Take

After doing this, IMO having the separate classes adds clarity to the taxonomy of errors; though we could do the same with a  set of typescript types; however the classes expose that taxonomy in raw javascript as well.

My concen with the single type model is that each user must mentally learn the categorization of errors based on properties in orcer to interact with them correctly - whereas having a taxonomy allows us to do this for them to some extent in a way they can leverage inspection.

But more comments and discussion welcome and expected.


cleve-fauna:
We're shipping with the hiearchy.

PR #41, opened 2023-02-14
Update query test assertions

cynicaljoy:
## Problem

Some of the tests are too explicit

## Solution

- Only assert the `data` on the result
- replace `.not.toBeUndefined()` with `toBeDefined()`
- don't validate `summary` contents, just that it's set 

## Result

A few tests pass

## Out of scope

Fix remaining tests

### DIFF
--- a/__tests__/functional/client-configuration.test.ts
+++ b/__tests__/functional/client-configuration.test.ts
@@ -43,7 +43,6 @@ an environmental variable named FAUNA_SECRET or pass it to the Client constructo
   });
 
   it("endpoints is extensible", async () => {
-    expect.assertions(4);
 
     endpoints["my-alternative-port"] = new URL("http://localhost:7443");
     expect(endpoints).toEqual({
--- a/__tests__/functional/client-configuration.test.ts
+++ b/__tests__/functional/client-configuration.test.ts
@@ -43,6 +43,8 @@ an environmental variable named FAUNA_SECRET or pass it to the Client constructo
   });
 
   it("endpoints is extensible", async () => {
+    expect.assertions(4);
+
     endpoints["my-alternative-port"] = new URL("http://localhost:7443");
     expect(endpoints).toEqual({
       cloud: new URL("https://db.fauna.com"),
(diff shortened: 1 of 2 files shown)


__tests__/functional/client-configuration.test.ts
```diff
@@ -60,7 +60,7 @@ an environmental variable named FAUNA_SECRET or pass it to the Client constructo
     expect(client.client.defaults.baseURL).toEqual("http://localhost:7443/");
     const result = await client.query<number>({ query: '"taco".length' });
     expect(result.txn_time).not.toBeUndefined();
-    expect(result).toEqual({ data: 4, txn_time: result.txn_time });
+    expect(result).toEqual({ data: 4, txn_time: result.txn_time, summary: "" });
```
cleve-fauna, line 63: This seems like odd behavior from Core.

@freels @macmv is it intended that core return empty strings for summary when there's no summary data?

freels: No, it probably shouldn’t do that. In general my philosophy too is that the drivers shouldn’t be testing core behavior like this, so if possible I’d tweak it to only check the data field itself or something.

cleve-fauna, line 63:
```suggestion
    expect(result).toMatchObject({ data: 4, txn_time: result.txn_time });
```

cleve-fauna: @cynicaljoy this will fix the test ^^^

cynicaljoy: I posted a fix before I saw this, let me know if you still want me to use this pattern or not. if not, please just mark conversation resolved.

cleve-fauna (approved):

__tests__/functional/client-configuration.test.ts
```diff
@@ -43,6 +43,8 @@ an environmental variable named FAUNA_SECRET or pass it to the Client constructo
   });
 
   it("endpoints is extensible", async () => {
+    expect.assertions(4);
```
cleve-fauna, line 46: You don't need this for tests that don't catch errors.

cleve-fauna, line 46:
```suggestion
```

cleve-fauna (changes requested):

__tests__/functional/client-configuration.test.ts
cleve-fauna: these are not needed for tests taht don't assertions in a catch block.

cleve-fauna (approved):

PR #42, opened 2023-02-14
GitHub Actions Workflow: second service

cynicaljoy:
## Problem

We have a test to tries a second local port

## Solution

Add a second `service` to the workflow

## Result

Test passes

## Out of scope

Fixing the remaining broken tests

### DIFF
--- a/.github/workflows/pr_validate.yml
+++ b/.github/workflows/pr_validate.yml
@@ -15,6 +15,10 @@ jobs:
         image: fauna/faunadb:latest
         ports:
           - 8443:8443
+      alt_core:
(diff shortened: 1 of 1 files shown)


cleve-fauna:
Which test is supposed to pass with this change @cynicaljoy ?

cynicaljoy:
> Which test is supposed to pass with this change @cynicaljoy ?

✅ [tests/functional/client-configuration.test.ts](https://github.com/fauna/fauna-js/runs/11337521216#r0s0)

ClientConfiguration
  ✅ endpoints is extensible

cleve-fauna (approved):

PR #43, opened 2023-02-14
Updated Query tests

cynicaljoy:
Ticket(s): FE-###

## Problem

❌ [tests/integration/query.test.ts](https://github.com/fauna/fauna-js/runs/11337521216#r0s3)

query with QueryRequest
  ❌ throws a QueryCheckError if the query is invalid
	Error: expect.assertions(5)
  ❌ throws a QueryCheckError if the query is invalid
	Error: expect.assertions(5)

## Solution

Fix them

## Result

✅

### DIFF
--- a/__tests__/integration/query.test.ts
+++ b/__tests__/integration/query.test.ts
@@ -179,7 +179,7 @@ describe.each`
     }
   });
 
-  it.skip("throws a QueryTimeoutError if the query times out", async () => {
+  it("throws a QueryTimeoutError if the query times out", async () => {
     expect.assertions(4);
     const badClient = new Client({
       endpoint: env["endpoint"] ? new URL(env["endpoint"]) : endpoints.local,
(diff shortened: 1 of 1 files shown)


__tests__/integration/query.test.ts
```diff
@@ -145,27 +143,20 @@ describe.each`
...
   it("throws a QueryCheckError if the query is invalid", async () => {
-    expect.assertions(5);
+    expect.assertions(4);
     try {
       await doQuery<number>(
         queryType,
-        getTsa`"taco".length;`,
-        '"taco".length;',
+        getTsa`happy little fox`,
+        "happy little fox",
```
cleve-fauna, line 151: what do you ahve against tacos?

cynicaljoy: it was valid FQL

cleve-fauna (approved):

__tests__/integration/query.test.ts
cleve-fauna: ah didn't used to be!

PR #44, opened 2023-02-14
Fix tsconfig warnings

cynicaljoy:
## Problem

image

## Solution

Listen to VS Code 🙈 

## Result

No more warnings 🎉 

## Out of scope

Doing real work


cleve-fauna:
doing the work the 🇺🇸 's won't do.

tsconfig.json
```diff
@@ -4,8 +4,10 @@
   "compilerOptions": {
     "lib": ["es2022"],
     "outDir": "./lib",
+    "declaration": true,
     "declarationDir": "./lib",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "ignoreDeprecations": "5.0"
```
cleve-fauna, line 10: what deprecations are we ignoring?

cleve-fauna: Oh I see from the screenshot.

My preference would be to get us in line with non-deprecated typescirpt. So we don't get a weird breakage down the line.

cynicaljoy: this seems to be caused by `"extends": "@tsconfig/node16-strictest/tsconfig.json",` but I'm not totally privy to if that's even necessary or what the node 18+ flavor would be. 

image

cleve-fauna: i'll set some time aside to help on this later today...

cynicaljoy: realized it would make more sense to bump from `@tsconfig/node16-strictest` to `@tsconfig/node18` I tried `@tsconfig/node18-strictest` but that was throwing warning/errors between vscode and jest that I couldn't get sorted out in a short amount of time.

cleve-fauna (approved):

PR #51, opened 2023-02-14
[FE 2983] Connection management

ptpaterson:
Ticket(s): [FE-2983](https://faunadb.atlassian.net/browse/FE-2983)

## Problem
The driver used Axios for alpha testing, but we need something that is
- compatible with many different environments.
- extensible for users (can bring your own fetch implementation)
- extensible for Fauna (can add custom HTTP2 support) 
- do all of this in a maintainable way

## Solution
Replace Axios with a thin wrapper over the Fetch API. 

To start, use the fetch API for all environments. This includes the Node fetch implementation (requires Node 18+). We can follow up and add different implementation with the `http2` module later.

## Result
The driver can be used anywhere that `fetch` can be used.

## Out of scope
HTTP2 support

## Testing
Update existing tests done with Axios so they work with `fetch`.

[FE-2983]: https://faunadb.atlassian.net/browse/FE-2983?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ptpaterson:
breaking this out into multiple PRs. starting with #58

PR #52, opened 2023-02-14
nitpick: Use a switch for the errors

cynicaljoy:
Ticket(s): FE-###

## Problem

It would be easier to read if we kept the status codes in order for the error codes

## Solution

Please ignore the changes in the `lib/*` files until we figure out what to do with those.

- Use a `switch`
- Removed `TODO` that applies to core, going to cut internal Jira card to track that effort

## Result

Improved readability


cleve-fauna (approved):
First you don't like tacos. Now you don't like huge if statements.

PR #55, opened 2023-02-15
Protect the secret

cynicaljoy:
Ticket(s): FE-3042

## Problem

We want to add some basic protections for the client `secret` to help ensure it's not accidentally logged

## Solution

Add some basic protections

## Result

Helping consumers keep their secrets safe 💪


cleve-fauna:
Without this change, what would a user have to do to splat their secret somewhere?

src/client.ts
```diff
@@ -59,6 +59,12 @@ export class Client {
       ...clientConfiguration,
       secret: this.#getSecret(clientConfiguration),
     };
+
+    // protect secret from being accidently logged
+    Object.defineProperty(this.clientConfiguration, "secret", {
+      enumerable: false,
+    });
```
cleve-fauna, line 66: without this block (but keeping removal from default headers) what would a user have to do to splat out their secret?

If they'd have to very intentionally log it I'm inclined to leave this alone so we don't get into making strange decisions about the objects people are passing into the client.

They might use that clientConfiguration elsehwere and mutating it this way under the covers could have unexpected consequences for them.

cleve-fauna: Another option would be to keep the secret in its own private variable, and keep the configuration in a separate variable with the secret omitted.

That would prevent us from mutating the user's input.

cynicaljoy: > without this block (but keeping removal from default headers) what would a user have to do to splat out their secret?

```
console.log(client)

    Client {
      clientConfiguration: {
        max_conns: 10,
        endpoint: URL {
          href: 'https://db.fauna.com/',
          origin: 'https://db.fauna.com',
          protocol: 'https:',
          username: '',
          password: '',
          host: 'db.fauna.com',
          hostname: 'db.fauna.com',
          port: '',
          pathname: '/',
          search: '',
          searchParams: URLSearchParams {},
          hash: ''
        },
        timeout_ms: 60000,
        secret: 'foo'
      },
      client: [Function: wrap] {
        constructor: [Function: wrap],
        request: [Function: wrap],
        getUri: [Function: wrap],
        delete: [Function: wrap],
        get: [Function: wrap],
        head: [Function: wrap],
        options: [Function: wrap],
        post: [Function: wrap],
        postForm: [Function: wrap],
        put: [Function: wrap],
        putForm: [Function: wrap],
        patch: [Function: wrap],
        patchForm: [Function: wrap],
        defaults: {
          transitional: [Object],
          adapter: [Function: httpAdapter],
          transformRequest: [Array],
          transformResponse: [Array],
          timeout: 70000,
          xsrfCookieName: 'XSRF-TOKEN',
          xsrfHeaderName: 'X-XSRF-TOKEN',
          maxContentLength: -1,
          maxBodyLength: -1,
          env: [Object],
          validateStatus: [Function: validateStatus],
          headers: [Object],
          baseURL: 'https://db.fauna.com/',
          httpAgent: [Agent],
          httpsAgent: [HttpsAgent]
        },
        interceptors: { request: [InterceptorManager], response: [InterceptorManager] },
        create: [Function: create]
      }
    }
```

I would actually like to use `enumerable: false` in combination w/changing `clientConfiguration` to be `protected` but we'll need to discuss the testing strategy.

cleve-fauna: i think there's other ways to acheive this. Let's huddle.

src/client.ts
```diff
@@ -56,42 +56,47 @@ export class Client {
...
       this.client.defaults.headers.common
     );
   }
 
+  /**
+   * Return the current {@link endpoint} for the Fauna Client
+   */
+  public get endpoint(): URL {
+    return this.#clientConfiguration.endpoint;
+  }
```
cleve-fauna, line 98: Let's just let them get the client configuration back.

```suggestion
  /**
   * Return the {@link ClientConfiguration} of this client, save for the secret.
   */
  public get clientConfiguration(): Omit<ClientConfiguration, "secret"> {
    const { secret, ...rest } = this.#clientConfiguration;
    return rest;
  }
```

cleve-fauna: See utility types: https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

cleve-fauna (approved):

PR #57, opened 2023-02-15
Client can set Last Txn Time

cynicaljoy:
Ticket(s): FE-2986

## Problem

Consumers should be able to update the Last Txn Time on their Client


__tests__/functional/client-configuration.test.ts
```diff
@@ -64,6 +63,27 @@ an environmental variable named FAUNA_SECRET or pass it to the Client constructo
...
+    const addFiveMinutes = new Date(expectedTxnTime.getTime() + 5 * 60000);
+    client.last_txn_time = addFiveMinutes;
+    expect(client.last_txn_time).toBe(addFiveMinutes);
+    // setting txn time back in history should fail
+    expect(() => {
+      client.last_txn_time = expectedTxnTime;
+    }).toThrow();
+    expect(() => {
+      client.last_txn_time = undefined;
+    }).toThrow();
```
cynicaljoy, line 84: I'm unsure about this bit, kinda feel like we should let customers "reset" it, but not sure if there's a reasonable use case for this or if we'd just expect them to init a new client. Question has been raised in Slack for broader input.

cynicaljoy: > in general, no. I would say it’s exceptional enough that they could in that case create a new client like you say

src/client.ts
```diff
@@ -90,6 +90,22 @@ export class Client {
...
+      if (time.valueOf() < this.last_txn_time.valueOf()) {
+        throw new Error("Must be greater than current value");
+      }
+    } else if (time == undefined && this.last_txn_time != undefined) {
+      throw new Error("Unable to clear last_txn_time");
+    }
+
+    this.#lastTxn = time;
+  }
+
```
cleve-fauna, line 108: I don't think we need to let them set this to undefined. Only to the future.

Also you don't need accessor modifiers in JS.

Additionally, let's stay consistent on snake vs camelCase in the field names of this class.

Finally, this is also the **reference driver** write good docs for all parts of the interface.

```suggestion
  /**
   * @return the last transaction time seen by this client, or undefined if this client has not seen a transaction time.
   */
  get lastTxn(): Date | undefined {
    return this.#lastTxn;
  }

   /**
   * Sets the last transaction time of this client.
   * @param lastTxn - the last transaction time to set.
   * @throws Error if lastTxn is before the current lastTxn of the driver
   */
  set lastTxnl(astTxn: Date) {
      if (this.lastTxn !== undefined && lastTxn.getTime() < this.lastTxn.getTime()) {
        throw new Error("Must be greater than current value");
      }
      this.#lastTxn = time;
  }
```

PR #58, opened 2023-02-15
[FE-3055] Update wire protocol to be consistent with latest openAPI spec

ptpaterson:
Ticket(s): [FE-3055](https://faunadb.atlassian.net/browse/FE-3055)

## Problem
types are out of date relative to the openAPI spec

## Solution/Result
Updated the types. Also updated the signatures for error constructors to use the types directly. 

## Testing
The signatures for Client methods `#getServiceError` and `#getError` have not changed, so the tests should only need updated to use the updated type names.

UPDATE: the existing tests made some incorrect assumptions about the wire protocol `summary` field, but those are also updated now.

[FE-3055]: https://faunadb.atlassian.net/browse/FE-3055?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/wire-protocol.d.ts
```diff
@@ -134,74 +154,41 @@ export declare class QueryTimeoutError extends ServiceError {
...
  * the request before the timeout was reached.
  */
 export declare class ServiceTimeoutError extends ServiceError {
-  constructor(error: {
-    code: string;
-    message: string;
-    httpStatus: 503;
-    summary?: string;
-  });
+  constructor(failure: QueryFailure, httpStatus: number);
```
cleve-fauna, line 191: we should enforce that `httpStatus` can only take the correct status code in these consturctors.

So replace `number` with `503`; same for all the other errors.

ptpaterson: got it. I was too aggressive with the copy-pasta. I'll fix that up.

lib/wire-protocol.d.ts
```diff
@@ -45,29 +45,64 @@ export interface QueryRequestHeaders {
...
-  stats: {
-    [key: string]: number;
+  /** The query's inferred static result type. */
+  static_type?: string;
+};
+/**
+ * A failed query response. Integrations which only want to report a human
+ * readable version of the failure can simply print out the "summary" field.
+ */
+export declare type QueryFailure = QueryInfo & {
```
cleve-fauna, line 89:
```suggestion
export declare type QueryError = QueryMetadata & {
```

lib/wire-protocol.d.ts
```diff
@@ -45,29 +45,64 @@ export interface QueryRequestHeaders {
...
-  /** A readable summary of any warnings or logs emitted by the query. */
-  summary?: string;
-}
+};
+export declare const queryResponseIsSuccess: <T>(
+  res: QueryFailure | QuerySuccess<T>
+) => res is QuerySuccess<T>;
+export declare const queryResponseIsFailure: (
+  res: QuerySuccess<any> | QueryFailure
+) => res is QueryFailure;
```
cleve-fauna, line 105:
```suggestion
export declare const isQueryResponse: <T>(
  res: QueryError | QueryResponse<T>
) => res is QueryResponse<T>;
export declare const isQueryError: (
  res: QueryResponse<any> | QueryError
) => res is QueryError;
```

lib/wire-protocol.d.ts
```diff
@@ -87,12 +122,7 @@ export declare class ServiceError extends Error {
    * where message does not suffice.
    */
   readonly summary?: string;
-  constructor(error: {
-    code: string;
-    message: string;
-    httpStatus: number;
-    summary?: string;
-  });
+  constructor(failure: QueryFailure, httpStatus: number);
```
cleve-fauna, line 125:
```suggestion
  constructor(failure: QueryError, httpStatus: number);
```

cleve-fauna (commented):
Have some naming suggestions.

Let me know what you think.

lib/wire-protocol.d.ts
```diff
@@ -45,29 +45,64 @@ export interface QueryRequestHeaders {
...
+  /** The query run time in milliseconds. */
+  query_time_ms: number;
+  /** The amount of data read from storage, in bytes. */
+  storage_bytes_read: number;
+  /** The amount of data written to storage, in bytes. */
+  storage_bytes_written: number;
+  /** The number of times the transaction was retried due to write contention. */
+  contention_retries: number;
+};
+export declare type QueryInfo = {
```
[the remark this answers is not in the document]
ptpaterson: `QueryInfo` comes from the openAPI spec. In general I am on board with `QueryMetadata` but I think it would be better to keep in sync with the openAPI spec.

lib/wire-protocol.d.ts
```diff
@@ -45,29 +45,64 @@ export interface QueryRequestHeaders {
...
   traceparent?: string;
 }
-/**
- * A response to a query.
- * @remarks
- * The QueryResponse is type parameterized so that you can treat it as a
- * a certain type if you are using Typescript.
- */
-export interface QueryResponse<T> {
+export declare type QueryInfoStats = {
```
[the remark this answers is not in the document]
ptpaterson: I am good with this.

lib/wire-protocol.d.ts
ptpaterson: `QueryFailure` comes from the openAPI spec too. So I've followed a convention of using "failure" for the whole payload. The error being the contents of the `error` field in that payload.

lib/wire-protocol.d.ts
ptpaterson: I like `isQuerySuccess` and `isQueryFailure` 👍🏻 

Or `isQueryResponse` and `isQueryError` if the respective types change.

ptpaterson: I've been thinking about it like

```typescript
type QueryResponse<T> = QuerySuccess<T> | QueryFailure
```

The `QuerySuccess<T> | QueryFailure` part is used in a few spots, so we could could declare the composed type, but I was thinking that being a little more verbose would be more clear that it we're handling either success or failure at that spot.

cleve-fauna: ah i see. yeah that makes sense.

lib/wire-protocol.d.ts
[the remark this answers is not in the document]
cleve-fauna: oh if that's from open API let's go with that.

lib/wire-protocol.d.ts
cleve-fauna: i see. yeah let's go with that then.

lib/wire-protocol.d.ts
ptpaterson: updated to isQuerySuccess and isQueryFailure

src/wire-protocol.ts
```diff
@@ -47,27 +47,69 @@ export interface QueryRequestHeaders {
...
+export type QueryFailure = QueryInfo & {
+  /**
+   * The result of the query resulting in
+   */
+  error: {
+    /** A predefined code which indicates the type of error. See XXX for a list of error codes. */
+    code: string;
+    /** description: A short, human readable description of the error */
+    message: string;
+  };
```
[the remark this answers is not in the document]
cleve-fauna: Moved to FE-3063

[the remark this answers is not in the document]
ptpaterson: Punting for now. Created an additional issue to come back to this.

src/wire-protocol.ts
```diff
@@ -47,27 +47,69 @@ export interface QueryRequestHeaders {
...
+};
+
+export const isQuerySuccess = <T>(
+  res: QuerySuccess<T> | QueryFailure
+): res is QuerySuccess<T> => "data" in res;
+
+export const isQueryFailure = (
+  res: QuerySuccess<any> | QueryFailure
+): res is QueryFailure => "error" in res;
```
cleve-fauna, line 112: we should document these and refine them - but they aren't used heavily until the fetch refactor so putning for now.

cleve-fauna (approved):

PR #61, opened 2023-02-16
[FE-3060] Implement fetch for node and web

ptpaterson:
Ticket(s): [FE-3060](https://faunadb.atlassian.net/browse/FE-3060)

## Problem
We need to get rid of Axios, because it does not work in a lot of environments

## Solution
Provide an initial implementation that just uses the native fetch API in browsers and Node. Add a custom fetch function to the client configuration that we can later provide various per-environment implementations (e.g. HTTP2 in Node), or that users can also provide.

## Result
Driver works with browsers and Node.

## Out of scope
HTTP2 node implementation

## Testing
Axios-specific tests need to be updated still.

[FE-3060]: https://faunadb.atlassian.net/browse/FE-3060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cleve-fauna:
@cynicaljoy @ptpaterson 

I'd recommend approaching implementation like this:

1. Have the goal of committing incremental, tested progress to main.
2. Build your interface as you go.

To acheive that:

1. Don't touch `Client` or the http layer client is using until you've made a new substitue for Axios.
2. Don't rip out axios until you've made its substitue.
3. In the meantime, build the substitutes in their own files. Test them as you go.
4. Once everything is tested replace the implementation in `Client` and remove axios.

The only time we should touch `Client` initially is if we want to make it accept *either* the axios or the substitute - so that you can start building end to end tests.

But doing that should not break the interface.

This will allow us to work more efficiently and collaborate more effectively while not blocking development on the main query / client interface.

PR #63, opened 2023-02-17
Test http client with mock fetch calls

ptpaterson:
Ticket(s): [FE-3071](https://faunadb.atlassian.net/browse/FE-3071)

Added some simple tests to check if the FetchClient is handling fetch requests correctly.

Used the `jest-fetch-mock` package, local to just the new test file. Thanks to @firehudson and @mmailaender for pointing to this package! It was quite easy to use.

[FE-3071]: https://faunadb.atlassian.net/browse/FE-3071?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/unit/fetch-client.test.ts
+++ b/__tests__/unit/fetch-client.test.ts
@@ -1,0 +1,113 @@
+import { enableFetchMocks } from "jest-fetch-mock";
+enableFetchMocks();
+
+import {
+  FetchClient,
+  HTTPRequest,
+  HTTPResponse,
+  isHTTPResponse,
+} from "../../src/http-client";
+import { QueryFailure, QuerySuccess } from "../../src/wire-protocol";
+
+let fetchClient: FetchClient;
+
+const dummyRequest: HTTPRequest = {
+  data: { query: "" },
+  headers: {},
+  method: "POST",
+  url: "",
+};
+
+const dummyStats = {
+  compute_ops: 0,
+  read_ops: 0,
+  write_ops: 0,
+  query_time_ms: 0,
+  storage_bytes_read: 0,
+  storage_bytes_written: 0,
+  contention_retries: 0,
+};
+
+describe("query", () => {
+  beforeAll(() => {
+    fetchClient = new FetchClient();
+  });
+
+  beforeEach(() => {
+    fetchMock.resetMocks();
+  });
+
+  it("returns a valid query response on success", async () => {
+    expect.assertions(7);
+    const expected: QuerySuccess<null> = {
+      data: null,
+      summary: "summary",
+      txn_time: "2023-02-17T19:19:12.887330Z",
+      query_tags: {},
+      stats: dummyStats,
+    };
+    const expectedHeaders = {
+      "content-type": "	application/json;charset=utf-8",
+      "x-faunadb-build": "230217.014809-a23359a",
+      "x-some-response-header": "Some header value",
+    };
+    fetchMock.mockResponseOnce(JSON.stringify(expected), {
+      headers: expectedHeaders,
+    });
+    const response: HTTPResponse = await fetchClient.request(dummyRequest);
+    if (isHTTPResponse(response)) {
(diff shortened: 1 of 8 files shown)


__tests__/unit/fetch-client.test.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  it("forwards errors when fetch rejects", async () => {
+    expect.assertions(1);
+    fetchMock.mockRejectOnce(new Error("oops"));
+    try {
+      await fetchClient.request(dummyRequest);
+    } catch (error) {
+      expect((error as Error).message).toEqual("oops");
+    }
+  });
```
ptpaterson, line 112: This has me thinking about error handling.

Right now the FetchClient just passes any errors on to the Client to deal with. This is moderately okay, because the implementation is so simple that pretty much the only real place it can fail is if there is if `fetch` rejects. The Client can assume any exceptions thrown are NetworkErrors.

I don't know if that is going to hold for other implementations. It seems likely that we will handle specific error cases, for example, with the Node HTTP2 implementation.

We probably want to enforce some contract that HTTPClients have a common base error to throw. We have several classes that extend the `Error` class. But we could create our own abstract `FaunaError` class from which all of our errors derive.

All that said, I think this test is reasonable, because all of our error classes make sure to construct the base Error class with the original failure message. E.g. we could catch and rethrow a different Error from the FetchClient and the test aught to continue to pass. We can add another test to check for the common Error class if we want.

ptpaterson: #64 provides that common error class. With that, I could update the fetch implementation to handle exceptions from `fetch` and add a test that it always returns an instance of the base class.

cleve-fauna: @ptpaterson let's punt this for now.

For discussion, I don't think we'd want the contract to be it throws a top-level error that is the parent of all errors - becuase if we end up making that error type we should make it abstract (that is never instantiate an actual member of it) because it doesn't really convey any information about what failed.

The concern would be that other HTTPClient implementations could get sloppy and emit a concrete version of an error that doesn't tell us what actually failed.

That being, said working on a contract for the the HTTP layer's request that includes what errors it fires sounds good - but let's punt it for the time being. I suspect it will be easier to work on a sane common contract once we have more than a single implementation.

__tests__/unit/fetch-client.test.ts
```diff
@@ -0,0 +1,113 @@
...
+    fetchMock.resetMocks();
+  });
+
+  it("returns a valid query response on success", async () => {
+    expect.assertions(7);
+    const expected: QuerySuccess<null> = {
+      data: null,
+      summary: "summary",
+      txn_time: "2023-02-17T19:19:12.887330Z",
+      query_tags: {},
```
cleve-fauna, line 46: minor nit

- let's make `query_tags` and `data` have some value.

That way the test verifies we're doing the work on the following lines:

https://github.com/fauna/fauna-js/blob/main/src/http-client/fetch-client.ts#L28

https://github.com/fauna/fauna-js/blob/main/src/http-client/fetch-client.ts#L30

cleve-fauna (commented):
Looks good!

Just have a minor suggestion to improve the happy case test.

__tests__/unit/fetch-client.test.ts
cleve-fauna: sounds good. If you want both go for it.

src/http-client/index.ts
```diff
@@ -8,7 +8,7 @@ export { FetchClient } from "./fetch-client";
  */
 export type HTTPRequest = {
   data: QueryRequest;
-  headers: Record<string, string>;
+  headers: Record<string, string | string[]>;
```
cleve-fauna, line 11: why is this an array?

cleve-fauna (approved):

PR #65, opened 2023-02-21
Tagged Type Formatting

cynicaljoy:
Ticket(s): FE-2979

## Problem

Latest version of FQL X defaults to tagged type formatting

## Solution

Implement tagged type formatting encoding/decoding

## Result

Can send/received tagged type format

## Out of scope

Updating the Client request/response to use `tagged` format

### DIFF
--- a/__tests__/functional/tagged-format.test.ts
+++ b/__tests__/functional/tagged-format.test.ts
@@ -89,50 +89,50 @@ describe("tagged format", () => {
     // expect(result).toEqual(
     //   `{"child":{"more":{"itsworking":{"@date":"1983-04-15T00:00:00.000Z"}}},"date":{"@date":"1923-05-13T00:00:00.000Z"},"decimal":{"@decimal":4.14},"long":{"@int":32},"name":"Hello, World","number":{"@int":48},"time":{"@time":"2023-01-30T21:27:45.204Z"}}`
     // );
-    const backToObj = JSON.parse(result);
+    const backToObj = JSON.parse(result)["@object"];
     expect(backToObj.decimal).toStrictEqual({ "@decimal": 4.14 });
     expect(backToObj.extra).toHaveLength(2);
     expect(backToObj.child.more.itsworking).toStrictEqual({
-      "@date": "1983-04-15T00:00:00.000Z",
+      "@date": "1983-04-15",
     });
   });
 
   it("handleds conflicts", () => {
     var result = TaggedTypeFormat.encode({
-      date: { "@date": new Date(2022, 11, 1) },
-      time: { "@time": new Date(2022, 11, 2) },
+      date: { "@date": new Date("2022-11-01T00:00:00.000Z") },
+      time: { "@time": new Date("2022-11-02T05:00:00.000Z") },
       int: { "@int": 1 },
       long: { "@long": 9999999999999 },
       double: { "@double": 1.99 },
     });
-    expect(result["date"]["@object"]["@date"]).toEqual(new Date(2022, 11, 1));
-    expect(result["time"]["@object"]["@time"]).toEqual(new Date(2022, 11, 2));
-    expect(result["int"]["@object"]["@int"]).toEqual(1);
-    expect(result["long"]["@object"]["@long"]).toEqual(9999999999999);
-    expect(result["double"]["@object"]["@double"]).toEqual(1.99);
+    expect(result["date"]["@object"]["@date"]).toStrictEqual({
+      "@date": "2022-11-01",
+    });
+    expect(result["time"]["@object"]["@time"]).toStrictEqual({
+      "@time": "2022-11-02T05:00:00.000Z",
+    });
+    expect(result["int"]["@object"]["@int"]).toStrictEqual({ "@int": 1 });
+    expect(result["long"]["@object"]["@long"]).toStrictEqual({
+      "@int": 9999999999999,
+    });
+    expect(result["double"]["@object"]["@double"]).toEqual({
+      "@decimal": 1.99,
+    });
   });
 
   it("handles nested conflict types", () => {
     expect(
       JSON.stringify(
         TaggedTypeFormat.encode({
-          "@object": {
+          "@date": {
             "@date": {
-              "@object": {
-                "@date": {
-                  "@object": {
-                    "@time": {
-                      "@time": "2022-12-02T02:00:00+00:00",
-                    },
-                  },
-                },
-              },
+              "@time": new Date("2022-12-02T02:00:00.000Z"),
             },
           },
         })
       )
     ).toEqual(
-      '{"@object":{"@date":{"@object":{"@date":{"@object":{"@time":{"@time":"2022-12-02T02:00:00+00:00"}}}}}}}'
+      '{"@object":{"@date":{"@object":{"@date":{"@object":{"@time":{"@time":"2022-12-02T02:00:00.000Z"}}}}}}}'
     );
   });
 
--- a/__tests__/functional/tagged-format.test.ts
+++ b/__tests__/functional/tagged-format.test.ts
@@ -86,9 +86,6 @@ describe("tagged format", () => {
       })
     );
 
-    // expect(result).toEqual(
-    //   `{"child":{"more":{"itsworking":{"@date":"1983-04-15T00:00:00.000Z"}}},"date":{"@date":"1923-05-13T00:00:00.000Z"},"decimal":{"@decimal":4.14},"long":{"@int":32},"name":"Hello, World","number":{"@int":48},"time":{"@time":"2023-01-30T21:27:45.204Z"}}`
-    // );
     const backToObj = JSON.parse(result)["@object"];
     expect(backToObj.double).toStrictEqual({ "@double": 4.14 });
     expect(backToObj.extra).toHaveLength(2);
--- a/__tests__/functional/tagged-format.test.ts
+++ b/__tests__/functional/tagged-format.test.ts
@@ -111,4 +111,28 @@ describe("tagged format", () => {
     expect(result["@long"]).toEqual(9999999999999);
     expect(result["@double"]).toEqual(1.99);
   });
+
+  it("handles nested conflict types", () => {
+    expect(
+      JSON.stringify(
+        TaggedTypeFormat.encode({
+          "@object": {
+            "@date": {
+              "@object": {
+                "@date": {
+                  "@object": {
+                    "@time": {
+                      "@time": "2022-12-02T02:00:00+00:00",
+                    },
+                  },
+                },
+              },
+            },
+          },
+        })
+      )
+    ).toEqual(
+      '{"@object":{"@date":{"@object":{"@date":{"@object":{"@time":{"@time":"2022-12-02T02:00:00+00:00"}}}}}}}'
+    );
+  });
 });
--- a/__tests__/functional/tagged-format.test.ts
+++ b/__tests__/functional/tagged-format.test.ts
@@ -65,7 +65,7 @@ describe("tagged format", () => {
       TaggedTypeFormat.encode({
         child: { more: { itsworking: new Date("1983-04-15") } },
         date: new Date("1923-05-13"),
-        decimal: 4.14,
+        double: 4.14,
         long: 32,
         name: "Hello, World",
         number: 48,
@@ -90,7 +90,7 @@ describe("tagged format", () => {
     //   `{"child":{"more":{"itsworking":{"@date":"1983-04-15T00:00:00.000Z"}}},"date":{"@date":"1923-05-13T00:00:00.000Z"},"decimal":{"@decimal":4.14},"long":{"@int":32},"name":"Hello, World","number":{"@int":48},"time":{"@time":"2023-01-30T21:27:45.204Z"}}`
     // );
     const backToObj = JSON.parse(result)["@object"];
-    expect(backToObj.decimal).toStrictEqual({ "@decimal": 4.14 });
+    expect(backToObj.double).toStrictEqual({ "@double": 4.14 });
     expect(backToObj.extra).toHaveLength(2);
     expect(backToObj.child.more.itsworking).toStrictEqual({
       "@date": "1983-04-15",
@@ -116,7 +116,7 @@ describe("tagged format", () => {
       "@int": 9999999999999,
     });
     expect(result["double"]["@object"]["@double"]).toEqual({
-      "@decimal": 1.99,
+      "@double": 1.99,
     });
   });
 
--- a/__tests__/functional/tagged-format.test.ts
+++ b/__tests__/functional/tagged-format.test.ts
@@ -12,7 +12,6 @@ describe("tagged format", () => {
       "name": "fir",
       "age": { "@int": "200" },
       "birthdate": { "@date": "1823-02-08" },
-      "molecules": { "@long": "999999999999999999" },
       "circumference": { "@double": "3.82" },
       "created_at": { "@time": "2003-02-08T13:28:12.000555+00:00" },
       "extras": {
@@ -34,7 +33,8 @@ describe("tagged format", () => {
           "employee": { "@int": "5" },
           "time": { "@time": "2023-02-08T14:22:01.000001+00:00" }
         }
-      ]
+      ],
+      "molecules": { "@long": "999999999999999999" }
     }`;
 
     const bugs_mod: Module = "Bugs";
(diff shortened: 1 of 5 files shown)


src/tagged-type.ts
```diff
@@ -0,0 +1,117 @@
...
+      } else if (value["@doc"]) {
+        return value["@doc"] as DocumentReference;
+      } else if (value["@int"]) {
+        return Number(value["@int"]);
+      } else if (value["@long"]) {
+        return Number(value["@long"]);
+      } else if (value["@decimal"]) {
+        return Number(value["@decimal"]);
+      } else if (value["@double"]) {
+        return Number(value["@double"]);
```
ptpaterson, line 37: Is it a problem that all of these values lose their types by the time you send them back to Fauna? Do we rely on type constructors in fqlx when stored type is important, e.g. `Int(${some_number})`?

This is not a concern when we have schema enforcement, so maybe that's the real answer here.

Javascript doesn't allow you to override operators, so the closest thing I think we could provide would be a class that implements `valueOf` so JS can coerce the value when it is used as a number. But then we are back to not having a type if you try to plug the result back into a query.

src/tagged-type.ts
```diff
@@ -0,0 +1,117 @@
...
+      } else if (value["@object"]) {
+        return value["@object"];
+      }
+
+      return value;
+    });
+  }
+}
+
+class taggedTypeEncoded {
```
ptpaterson, line 51:
```suggestion
class TaggedTypeEncoded {
```

Capitalize the class name

cynicaljoy: Go habit of non-exported objects be lower case ;)

src/tagged-type.ts
cynicaljoy: Yeah I'm not quite sure how to handle this with JavaScript since it only has the one type. Maybe we need to provide our own types like we're doing for `Module` and `DocumentReference`.

ptpaterson: I mostly wanted to spark the conversation and thoughts from others. It is reasonable to me to move forward decoding to `Number` at this time.

__tests__/functional/tagged-format.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      "circumference": { "@double": "3.82" },
+      "created_at": { "@time": "2003-02-08T13:28:12.000555+00:00" },
+      "extras": {
+        "nest": {
+          "@object": {
+            "num_sticks": { "@int": "58" },
+            "@extras": { "egg": { "fertilized": false } }
+          }
+        }
+      },
```
ptpaterson, line 25: In this part of the test, `@object` indicates a tagged object and `@extras` is literally the field name. since `@extras` is an object, then it should also include an `@object` field, as should `egg`

```suggestion
      "extras": {
        "nest": {
          "@object": {
            "num_sticks": { "@int": "58" },
            "@extras": { 
              "@object": {
                "egg": { 
                  "@object": { "fertilized": false } 
                }
              } 
            }
          }
        }
      },
```

src/tagged-type.ts
```diff
@@ -0,0 +1,117 @@
...
+      } else if (value["@decimal"]) {
+        return Number(value["@decimal"]);
+      } else if (value["@double"]) {
+        return Number(value["@double"]);
+      } else if (value["@date"]) {
+        return new Date(value["@date"]);
+      } else if (value["@time"]) {
+        return new Date(value["@time"]);
+      } else if (value["@object"]) {
+        return value["@object"];
```
ptpaterson, line 43: I think this should call decode recursively

src/tagged-type.ts
```diff
@@ -0,0 +1,117 @@
...
+      return value;
+    });
+  }
+}
+
+class TaggedTypeEncoded {
+  readonly result: any;
+
+  readonly #encodeMap = {
+    number: (value: number): { [key: string]: number } => {
```
ptpaterson, line 55: we can be more specific here:

```suggestion
    number: (value: number): { '@int':  number } | { '@decimal': number } => {
```

src/tagged-type.ts
```diff
@@ -0,0 +1,117 @@
...
+      const _out: { [key: string]: any } = {};
+      for (const k in input) {
+        if (k.startsWith("@")) {
+          _out[k] = input[k];
+        } else {
+          _out[k] = TaggedTypeFormat.encode(input[k]);
+        }
+      }
+      return _out;
+    },
```
ptpaterson, line 75: This I think needs some work.

- the result should be wrapped in an `"@object"` field. 
- fields provided from the user, starting with `@` only matter if the field is an object and contains exactly one field starting with `@`. If there are multiple fields starting with `@` we can interpret the object as having user-defined fields that happen to start with `@`

__tests__/functional/tagged-format.test.ts
cynicaljoy: from my understanding `@object` is optional. from local testing it seems that nesting the contents of `egg` doesn't change the result any. the object structure was copied directly from what we're using to [validate Python driver](https://github.com/fauna/fauna-python/blob/4ac342a22aaf9d30b3175d3e80c4c7a410c69ee7/tests/fixtures/all_types.py#L5-L11) -- so if we feel strongly that we _need_ this then we should update in both places. However, I don't see that it's necessary.

src/tagged-type.ts
cynicaljoy: > reviver A function that transforms the results. This function is called for each member of the object.
> If a member contains nested objects, the nested objects are transformed before the parent object is.

shouldn't be necessary

src/tagged-type.ts
cynicaljoy: I don't think this adds any value as this is just an internal func

ptpaterson: And a really small internal function at that. Just a nitpick.

src/tagged-type.ts
cynicaljoy: > We can't assume that fields starting with @ are Fauna tagged fields.

We're not. Per the [spec](https://github.com/fauna/core/blob/master/ext/api/src/main/resources/openapi.yml#L442-L445):
> If an FQL object value contains properties starting with `@`, it will be emitted on the wire in an `@object` tag.

It one of the fields of their object has an `@` prefix, we just pass it through as an object rather than applying any tagging to it.

ptpaterson: For example `{ "@foo": new Date("2022-12-02") }` should encode to 

```JSON
{
  "@object": {
    "@foo": {
      "@date": "2022-12-02"
    }
  }
}
```

but will currently encode to 

```JSON
{
  "@foo": "2022-12-01T05:00:00.000Z"
}
```

ptpaterson: So, the `@object` tag is not strictly necessary when there are no other "@" fields, but it is necessary when there are tagged types. It might be easier if the driver just always wraps in `@object`

we can provide an `foo` object field without the `@object` wrapper if there are no user-defined "@" fields

image

if we rename the field to `@foo` we get an error.

image

but we can fix it by wrapping in `@object`

image

__tests__/functional/tagged-format.test.ts
```diff
@@ -0,0 +1,138 @@
...
+              },
+            },
+          },
+        })
+      )
+    ).toEqual(
+      '{"@object":{"@date":{"@object":{"@date":{"@object":{"@time":{"@time":"2022-12-02T02:00:00+00:00"}}}}}}}'
+    );
+  });
+});
```
ptpaterson, line 138:
```suggestion
  });
  
  it("wraps user-provided `@` fields", () => {
    expect(
      JSON.stringify(
        TaggedTypeFormat.encode({
          "@foo": true,
        })
      )
    ).toEqual('{"@object":{"@foo":true}}');
  });
});
```

src/tagged-type.ts
cynicaljoy: I see what you're saying now, thanks for the screenshots w/the curl examples. @pnwpedro given this we should update the Python driver encoding too.

__tests__/functional/tagged-format.test.ts
```diff
@@ -99,17 +99,17 @@ describe("tagged format", () => {
 
   it("handleds conflicts", () => {
     var result = TaggedTypeFormat.encode({
-      "@date": new Date(2022, 11, 1),
-      "@time": new Date(2022, 11, 2),
-      "@int": 1,
-      "@long": 9999999999999,
-      "@double": 1.99,
+      date: { "@date": new Date(2022, 11, 1) },
```
ptpaterson, line 102: here, `"@date"` is supposed to be a user-field, right? But then we assign to it a `Date`. Should the `Date` still get encoded as `{ "@date": "2022-11-01" }`?

I think the test should then be

```
expect(result["date"]["@object"]["@date"]).toEqual({"@date": "2022-11-01"});
```

src/tagged-type.ts
```diff
@@ -66,7 +66,7 @@ class TaggedTypeEncoded {
       const _out: { [key: string]: any } = {};
       for (const k in input) {
         if (k.startsWith("@")) {
-          _out[k] = input[k];
+          _out["@object"] = k == "@object" ? input[k] : { [k]: input[k] };
```
ptpaterson, line 69: We still need to call `TaggedTypeFormat.encode` on the value if it's a user-defined field.

__tests__/functional/tagged-format.test.ts
```diff
@@ -0,0 +1,148 @@
...
+          {
+            id: 1,
+            time: new Date(),
+          },
+          {
+            id: 2,
+            time: new Date(),
+          },
+        ],
+        "@foobar": {
```
ptpaterson, line 83: since the object to be encoded includes a field called `@foobar`, which starts with `@`, then the whole object will need to be wrapped in `@object`. That breaks all the tests that follow.

src/tagged-type.ts
pnwpedro: > Maybe we need to provide our own types like we're doing for Module and DocumentReference.

I see this convo is resolved, but this is what popped into my mind, as well. Did you discuss elsewhere?

src/tagged-type.ts
pnwpedro: Does [this example](https://github.com/fauna/fauna-python/blob/dev/tests/fixtures/tag_conflicts.py#L56-L80) in the python driver mean we're covered?

src/tagged-type.ts
ptpaterson: No just resolved to move forward with this PR.

If we made our own JS classes for things like `FaunaInt`, `FaunaDecimal` and even `FaunaDate` and `FaunaTime` (the latter of which we have in the v4 driver), then I don't know how useful it would be to _decode_ responses into these classes. 

As I was working later on testing, I realized how persnickety it is to encode `Date` vs `Time` types. i.e. a JS `Date` is only a Fauna `Date` if it's at 00:00:00 UTC. It would be nice for users to cast use a JS `Date` as a Fauna `Date` regardless of the time.

That makes me think back to numbers. We've had users want to store floating point values that happen to be whole numbers. So it would be nice to have a mechanism to explictly cast numbers as either Fauna ints or floats.

Now, folks can cast things now using FQL constructors, but that is limited to the parts that you handle directly in FQL. Sometimes you have an object in code you just want to store, e.g. `somedoc.update(${data})` where data contains fields that should be cast. Passing around explicit Fauna types could help with that.

I would support a fast follow up to add date and time classes and decode responses to those instead of immediately to Date.

If would also support classes for int, long, decimal & double. However, I am not convinced we should decode directly to those, but rather we could have them as a mechanism for users to ensure their values get encoded in a consistent way.

src/tagged-type.ts
```diff
@@ -0,0 +1,118 @@
...
+  static decode(input: string): any {
+    return JSON.parse(input, (_, value: any) => {
+      if (value["@mod"]) {
+        return value["@mod"] as Module;
+      } else if (value["@doc"]) {
+        return value["@doc"] as DocumentReference;
+      } else if (value["@int"]) {
+        return Number(value["@int"]);
+      } else if (value["@long"]) {
+        return Number(value["@long"]);
```
pnwpedro, line 33: I don't think this is safe. JS doesn't support 64 bit integers because all numbers are 64 bit floats in JS. We prob need to branch further if we're not using fauna-vended types, but i'm starting to lean toward fauna-vended types as a way to guarantee consistency and mitigate data loss/confusion.

We talk about this in our docs in case that helps. https://docs.fauna.com/fauna/current/build/fql/types?lang=javascript#large-integers

src/tagged-type.ts
```diff
@@ -0,0 +1,118 @@
...
+      } else if (value["@decimal"]) {
+        return Number(value["@decimal"]);
```
pnwpedro, line 35: I didn't see this as part of the wire protocol. Is it a change I missed?

src/tagged-type.ts
pnwpedro: We should probably make a decision on this. Post scrum topic today!

src/tagged-type.ts
ptpaterson: @pnwpedro  I agree. Python and JS should be in sync an both covering , considering the test you linked to and [the encoding for dicts](https://github.com/fauna/fauna-python/blob/0dfee551c312f0689d08e9bacc42349e6c75c00b/fauna/encode.py#L68-L72) lines up with JS, too.

pnwpedro (changes requested):

src/tagged-type.ts
pnwpedro: We should probably make a decision on this. Post scrum topic today!

pnwpedro: > If would also support classes for int, long, decimal & double. However, I am not convinced we should decode directly to those, but rather we could have them as a mechanism for users to ensure their values get encoded in a consistent way.

regarding this, the problem is that we need to guarantee safe round-tripping. we'll have to reach consensus, but i think of this as a hard requirement

src/tagged-type.ts
cynicaljoy: yeah I think we need to vend types, I should have put a TODO comment here as I was unsure of what to do but wanted to keep rolling.

src/tagged-type.ts
cynicaljoy: no, I just went nutty with types

src/tagged-type.ts
cleve-fauna: modern javascript has a BigInteger type.

IMO we should use that.

src/tagged-type.ts
```diff
@@ -0,0 +1,116 @@
...
+      } else if (value["@int"]) {
+        return Number(value["@int"]);
+      } else if (value["@long"]) {
+        return Number(value["@long"]);
+      } else if (value["@double"]) {
+        return Number(value["@double"]);
+      } else if (value["@date"]) {
+        return new Date(value["@date"] + "T00:00:00.000Z");
+      } else if (value["@time"]) {
+        return new Date(value["@time"]);
```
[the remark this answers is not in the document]
cynicaljoy: that returns a number rather than a Date. So you want me to use `new Date(Date.parse(value["@date"]))`?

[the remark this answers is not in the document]
cleve-fauna: oh yuck. nevermind i was wrong about that.

src/tagged-type.ts
ptpaterson: Added handling of Fauna Longs as BigInt

src/tagged-type.ts
ptpaterson: Number handling is updated using `number` and `BigInt` without any custom classes. I've stubbed out classes and notes for `FaunaDate` and `FaunaTime`. We will model after the TC39 proposed `Temporal.PlainDate` and `Temporal.Instant`, respectively, with the intent to replace these implementations with the Temporal types when they are stable.

pnwpedro (approved):

src/tagged-type.ts
```diff
@@ -0,0 +1,141 @@
...
+          "@long": value.toString(),
+        };
+      }
+      throw new TypeError("Precision loss when converting int to Fauna type");
+    },
+    number: (value: number): TaggedDouble | TaggedInt | TaggedLong => {
+      if (`${value}`.includes(".")) {
+        return { "@double": value };
+      } else {
+        if (value >= -(2 ^ 31) + 1 && value <= 2 ** 31 - 1) {
```
cleve-fauna, line 72: `^` is the XOR operator. This looks like a bug.

Also, 2's complement is [- 2^(n-1), 2^(n-1) - 1]. Our range is incorrect.

We are also missing tests that would have caught these issues.

Let's add some basic tests to test number ranges and mappings.

PR #66, opened 2023-02-22
Replace Axios

ptpaterson:
Ticket(s): [FE-3060](https://faunadb.atlassian.net/browse/FE-3060)

## Problem
We need to get rid of Axios, because it does not work in a lot of environments

## Solution
Provide an initial implementation that just uses the native fetch API in browsers and Node. Add a custom fetch function to the client configuration that we can later provide various per-environment implementations (e.g. HTTP2 in Node), or that users can also provide.

## Result
Driver works with browsers and Node.

## Out of scope
- HTTP2 node implementation
- Tagged type encoding, but I added `// WIP` comments where it should be straightforward to drop that in when we're ready.

## Testing
Tests updated to reflect new API. Notes:
- deleted:    `__tests__/integration/connection-pool.test.ts`
- commented out some tests in `__tests__/unit/query.test.ts`. We should review and either delete them or determine their analog for the new HTTPClient.

[FE-3060]: https://faunadb.atlassian.net/browse/FE-3060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client.ts
```diff
@@ -34,14 +39,15 @@ const defaultClientConfiguration = {
 export class Client {
   /** The {@link ClientConfiguration} */
   readonly #clientConfiguration: ClientConfiguration;
-  /** The underlying {@link AxiosInstance} client. */
-  readonly client: AxiosInstance;
+  /** The underlying {@link HTTPClient} client. */
+  readonly client: HTTPClient;
```
ptpaterson, line 43: We agreed to rename this `#httpClient`, to 1) distinguish from a "client" as in an instance of the `Client` class, and 2) to make it private.

cleve-fauna: any reason we can't make that change now?

src/client.ts
```diff
@@ -34,14 +39,15 @@ const defaultClientConfiguration = {
...
   /** last_txn this client has seen */
   #lastTxn?: Date;
 
   /**
    * Constructs a new {@link Client}.
    * @param clientConfiguration - the {@link ClientConfiguration} to apply.
+   * @param client - The underlying {@link HTTPClient} that will execute the actual HTTP calls.
```
cleve-fauna, line 50:
```suggestion
   * @param clientConfiguration - the {@link ClientConfiguration} to apply. Defaults to recommended ClientConfiguraiton.
   * @param client - The underlying {@link HTTPClient} that will execute the actual HTTP calls. Defaults to recommended HTTPClient.
```

src/client.ts
ptpaterson: Yea, I intended to resolve this now.

src/client.ts
```diff
@@ -155,36 +143,33 @@ export class Client {
...
-        const failure = e.response.data;
-        const status = e.response.status;
+      if (isQueryFailure(e.body)) {
+        const failure = e.body;
+        const status = e.status;
         return this.#getServiceError(failure, status);
       }
+
       // we got a different error from the protocol layer
       return new ProtocolError({
```
cleve-fauna, line 167: we seem to be losing the distinction between protocol and network errors.

This is a bit troubling as network errors are usually retriable and in P2 we'd like to build retries on these.

Can we use the guidance here: https://fetch.spec.whatwg.org/#concept-network-error to correctly determine a network error?

src/client.ts
```diff
@@ -239,27 +224,66 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+        parsedResponse = {
+          ...fetchResponse,
+          // WIP: add decoding here
+          body: JSON.parse(fetchResponse.body),
+        };
+      } catch (error: unknown) {
+        throw new ProtocolError({
+          message: `Error parsing response as JSON: ${error}`,
+          httpStatus: fetchResponse.status,
+        });
```
cleve-fauna, line 267: this will always throw a [SyntaxError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#exceptions) if this is not valid JSON.

Let's do a couple things:

1. remove this catch block
2. update `#getError` to map `SyntaxError` to a protocol error.

That will do a couple things:

1. `#getError` remains the only place responsible for producing errors
2. We tighten the rule on what gets lumped as a protocol error for invalid json.

src/http-client/fetch-client.ts
```diff
@@ -20,6 +21,10 @@ export class FetchClient implements HTTPClient {
       method,
       headers: { ...requestHeaders, "Content-Type": "application/json" },
       body: JSON.stringify(data),
+    }).catch((error) => {
+      throw new NetworkError("The network connection encountered a problem.", {
+        cause: error,
+      });
     });
```
cleve-fauna, line 28: this is too generic.

There's no guarantee this a Network error. See [fetch docs](https://developer.mozilla.org/en-US/docs/Web/API/fetch#exceptions) for other possiblities.

Let's delete this block and let the client class sort it out.

src/http-client/index.ts
```diff
@@ -8,7 +9,7 @@ export { FetchClient } from "./fetch-client";
  */
 export type HTTPRequest = {
   data: QueryRequest;
-  headers: Record<string, string | string[]>;
+  headers: Record<string, string>;
```
cleve-fauna, line 12: is this revert intentional?

src/client.ts
ptpaterson: done!

src/http-client/index.ts
ptpaterson: Yea, I applied the change before to request headers, but I mean to only update the headers type for response headers.

src/http-client/fetch-client.ts
ptpaterson: Ah, makes sense. I had reviewed the fetch docs, but I think got hung up on this line at the top

> A [fetch()](https://developer.mozilla.org/en-US/docs/Web/API/fetch) promise only rejects when a network error is encountered (which is usually when there's a permissions issue or similar).

So once the request is made, it's definitely a network error, but while you are generating a request it can still throw other errors 👍🏻 

We can check for `error instance of DOMException` and `error.name === "NetworkError"`.

Note that `DOMException` only makes sense in the context of web APIs (e.g. fetch). That's probably more than the Client should have to worry about. Counter argument is that since developers can provide their own HTTPClient we would have better error handling if the developer fails to handles the errors themselves. Favoring better error handling in the user-provided http client, I'll push the error handling to the Client class.

src/client.ts
ptpaterson: One reason this is handled inline with the httpRequest is that the status code is in scope when constructing the ProtocolError. We lose that if we just let the try-catch send the error to `#getError`.

So I am thinking about how we can have a more specific contract with `#getError` so that we know if the argument is pre-request or post-request along with the appropriate context.

src/client.ts
ptpaterson: This goes back to me thinking of network errors having to be handled in the HTTPClient, where they have context about what state the request is in. In that case, `#getError` is not handling any network errors.

The spec link says all of the reasons that a request should be a network error, but it doesn't include the implementation. The promise returned from `fetch` will already reject in those cases, so we don't have to check for status code == 0, etc. 

I have to figure out how to always pass along the appropriate context to `#getError` if we're punting everything to the client. Axios errors conveniently provided that context, but `fetch` does not.

src/http-client/fetch-client.ts
ptpaterson: The distinction is between the Promise for `fetch()` rejecting and the Promise for `response.text()` rejecting. The former indicates one of several errors like bad headers, for example. The latter will always be a network error.

It is straightforward to make the distinction inside of the HTTPClient. both the bad-request errors and network errors are thrown as `TypeError` we cannot make the distinction outside of the HTTPClient.

I think we should update FetchClient to catch and re-throw a `NetworkError` on `response.text()` and catch and re-throw a `ProtocolError` on the `fetch()` call.

cleve-fauna: bad headers do not fit with our definition of network error. we should huddle on this to get on the same page for terminology.

src/client.ts
```diff
@@ -34,15 +39,15 @@ const defaultClientConfiguration = {
...
+  readonly #httpClient: HTTPClient;
   /** last_txn this client has seen */
   #lastTxn?: Date;
 
   /**
    * Constructs a new {@link Client}.
-   * @param clientConfiguration - the {@link ClientConfiguration} to apply.
-   * @example
+   * @param clientConfiguration - the {@link ClientConfiguration} to apply. Defaults to recommended ClientConfiguraiton.
+   * @param httpClient - The underlying {@link HTTPClient} that will execute the actual HTTP calls. Defaults to recommended HTTPClient.   * @example
```
cleve-fauna, line 50:
```suggestion
   * @param httpClient - The underlying {@link HTTPClient} that will execute the actual HTTP calls. Defaults to recommended HTTPClient.   
   * @example
```

src/client.ts
cleve-fauna: discussed offline. moving some of the distictions to a P2.

src/http-client/fetch-client.ts
cleve-fauna: discussed offline. moving finer distinctions to P2

src/client.ts
cleve-fauna: discussed offline. moving finer distinctions to P2

cleve-fauna (approved):

PR #67, opened 2023-02-22
add simple integ test against existing collection

fauna-chase:
We have now finalized our collection data format and will begin giving some customers early access to fql x.  This test is going to ensure that we don't make any backwards incompatible changes to our collection data format.  The plan is to run this after the dev and preview deploys in the database pipeline.

We also plan to run the full drivers test suite in these pipelines to ensure we don't have any backwards incompatible drivers changes.  Some of those tests don't currently work when pointed at dev, taking an incremental approach here and will be getting this test in the pipeline first while I work on resolving the other drivers tests.  At that point will switch to running all of the integ tests in the pipeline.

### DIFF
--- a/__tests__/integration/existing-collection.test.ts
+++ b/__tests__/integration/existing-collection.test.ts
@@ -1,0 +1,119 @@
+import { Client } from "../../src";
+import { endpoints } from "../../src";
+import { ServiceError } from "../../src";
+import { env } from "process";
+import { fql } from "../../src";
+
+const client = new Client({
+  endpoint: env["endpoint"] ? new URL(env["endpoint"]) : endpoints.local,
+  max_conns: 5,
+  secret: env["secret"] || "secret",
+});
+
+const authorsData = [
+  {
+    firstName: "one",
+    lastName: "one",
+    genre: "mystery",
+  },
+  {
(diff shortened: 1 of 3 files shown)


adambollen (dismissed):

__tests__/integration/existing-collection.test.ts
```diff
@@ -0,0 +1,119 @@
...
+import { endpoints } from "../../src";
+import { ServiceError } from "../../src";
+import { env } from "process";
+import { fql } from "../../src";
+
+const client = new Client({
+  endpoint: env["endpoint"] ? new URL(env["endpoint"]) : endpoints.local,
+  max_conns: 5,
+  secret: env["secret"] || "secret",
+});
```
cleve-fauna, line 11: We've got a test helper function for this.

Note, it uses "FAUNA_SECRET" and "FAUNA_ENDPOINT" as the env vars.

```suggestion
import { getClient } from "../client";
import { ServiceError } from "../../src";
import { fql } from "../../src";

const client = getClient();
```

cleve-fauna (commented):
Looks good!

Thanks for this.

One small suggestion.

adambollen (approved):

PR #70, opened 2023-02-24
Fix bugs in typed tag translations, add more tests for numbers

cleve-fauna:
currently this is a work in progress

Ticket(s): FE-2979 

## Problem
- bugs in typed tag translations for numbers
## Solution
- fix bugs and add tests that exercise number cases
## Result
- can translate numbers correctly end to end

## Testing
- full suite - tests currently fail. need to get more cycles on this.


ptpaterson:
This looks good to me. The remaining tests fail because encoding/decoding is not actually hooked up to the client.

ptpaterson (approved):

PR #71, opened 2023-02-24
Use tagged format

ptpaterson:
Ticket(s): [FE-3081](https://faunadb.atlassian.net/browse/FE-3081)

## Problem
We implemented the HTTP client and the tagged format. Now we need to hook them up.

## Result
The client defaults to using the "tagged" format, but users can configure the client with either "simple" or "tagged", and format can be configured per query as well.

## Testing
Added a test to make sure configuring format sends the right header. Fixed tests handling BigInt.

[FE-3081]: https://faunadb.atlassian.net/browse/FE-3081?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


__tests__/functional/tagged-format.test.ts
```diff
@@ -161,15 +161,15 @@ describe("tagged format", () => {
...
-    ${-(2 ** 63)}     | ${BigInt(-(2 ** 63))}     | ${"bigint"}  | ${"-2**63"}
-    ${2 ** 31 - 1}    | ${2 ** 31 - 1}            | ${"number"}  | ${"2**31 - 1"}
-    ${2 ** 31}        | ${BigInt(2 ** 31)}        | ${"bigint"}  | ${"2**31"}
-    ${2 ** 63 - 1}    | ${BigInt(2 ** 63 - 1)}    | ${"bigint"}  | ${"2**63 - 1"}
-    ${1.3 ** 63}      | ${1.3 ** 63}              | ${"number"}  | ${"1.3**63"}
-    ${1.3}            | ${1.3}                    | ${"number"}  | ${"1.3"}
+    input                             | expected                          | expectedType | testCase
+    ${-(2 ** 31)}                     | ${-(2 ** 31)}                     | ${"number"}  | ${"-2**31"}
+    ${-(2 ** 31) - 1}                 | ${BigInt(-(2 ** 31) - 1)}         | ${"bigint"}  | ${"-2**31 - 1"}
+    ${BigInt("-9223372036854775808")} | ${BigInt("-9223372036854775808")} | ${"bigint"}  | ${"-2**63"}
```
cleve-fauna, line 167: i had this as a number intentionally as we seem to take a number and send it to Fauna as a long when its over 32 bits. It would then come out as a BigInt.

This looks likea  good additional case too.

But if my reading of the serializer / deserializer is correct we should keep that behavior as well.

Alghouth my tests was a bit crap since JS can't go over ~ 2 ^ 53 for numbers.

So maybe we add a case here for something like 2 ^ 40 and - (2 ^ 40) ?

ptpaterson: the number was too big for a fauna number, that's why I wrapped the input in BigInt.

ptpaterson: If the JS number is too big for an int, then it will become a double.

cleve-fauna: That's not how the code reads:

https://github.com/fauna/fauna-js/blob/main/src/tagged-type.ts#L77-L79

and now that i'm looking again i'm not sure this code is safe. As we are calling 2 ^ 63 which exceeds: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

cleve-fauna: > the number was too big for a fauna number, that's why I wrapped the input in BigInt.

Please re-read my comment. I see that. I am suggesting an additional test.

ptpaterson: sorry I meant too big for long

cleve-fauna: there's still some bugs in the source let's get those cleaned up and then we can cover the test cases

ptpaterson: okay, the problem was I saw that the int was too big to really be an int in JS so I was agressive about making it a BigInt. It should still work with the driver turning the JS number into a BigInt. No more argument, let's add that back.

ptpaterson: test appears to be failing for trying to compare BigInts with Jest `.toEqual` function. If I comment that out it passes.

cleve-fauna: yeah but again i think we have a bug in the code.

The max safe int i is 2 ^ 53; and we are blowing past that in our source code.

So let's fix that up and also fix jest.

ptpaterson: comparing with .toString allows the tests to run, but of course passing in a js number too precise for a float truncates the value so we still need to adjust the test.

cleve-fauna (changes requested):
Please fix the bug that is using unsafe ints in the serializer.

__tests__/functional/tagged-format.test.ts
ptpaterson: Okay, we're there! `Math.isSafeInteger` to the rescue!! 🙏🏻

ptpaterson: Actually I am working on one more test case that throws the TypeError when the BigInt would lose precision.

ptpaterson: That's done, too.

src/tagged-type.ts
```diff
@@ -70,16 +73,16 @@ class TaggedTypeEncoded {
     },
     number: (value: number): TaggedDouble | TaggedInt | TaggedLong => {
       if (`${value}`.includes(".")) {
-        return { "@double": value };
+        return { "@double": value.toString() };
```
cleve-fauna, line 76: This is a bug:

```
> `${2.1**70}`
'3.5921183035610026e+22'
```

Do you have to transmit it as a string? If so, we need to deliver a string correctly and not in this JS format.

src/tagged-type.ts
```diff
@@ -70,16 +73,16 @@ class TaggedTypeEncoded {
...
-          return { "@int": value };
-        } else if (value >= -(2 ** 63) && value <= 2 ** 63 - 1) {
+          return { "@int": value.toString() };
+        } else if (Number.isSafeInteger(value)) {
           return {
             "@long": value.toString(),
           };
         }
-        return { "@double": value };
+        return { "@double": value.toString() };
```
cleve-fauna, line 85: This is a bug:

```
> 2**100
1.2676506002282294e+30
```

Note also it will suffer the same precision loss as line 72.

We should be consistent.

Either:

1. reject any precision loss
2. mutate places we lose precision to doubles; and only blow up if we go past Fauna or JS's max double size.

IMO (2) is preferable. Though we need to account for precision loss with documentation or warnings to users.

cleve-fauna (dismissed):
Still a couple bugs.

__tests__/functional/tagged-format.test.ts
```diff
@@ -161,22 +164,49 @@ describe("tagged format", () => {
...
+    ${Number.MIN_SAFE_INTEGER}     | ${BigInt(Number.MIN_SAFE_INTEGER)} | ${"bigint"}  | ${"@long"}   | ${"-(2**53 - 1)"}
+    ${-(2 ** 31) - 1}              | ${BigInt(-(2 ** 31) - 1)}          | ${"bigint"}  | ${"@long"}   | ${"-(2**31) - 1"}
+    ${-(2 ** 31)}                  | ${-(2 ** 31)}                      | ${"number"}  | ${"@int"}    | ${"-(2**31)"}
+    ${0}                           | ${0}                               | ${"number"}  | ${"@int"}    | ${"0 (Int)"}
+    ${1}                           | ${1}                               | ${"number"}  | ${"@int"}    | ${"1 (Int)"}
+    ${BigInt("0")}                 | ${BigInt("0")}                     | ${"bigint"}  | ${"@long"}   | ${"0 (Long)"}
+    ${2 ** 31 - 1}                 | ${2 ** 31 - 1}                     | ${"number"}  | ${"@int"}    | ${"2**31 - 1"}
+    ${2 ** 31}                     | ${BigInt(2 ** 31)}                 | ${"bigint"}  | ${"@long"}   | ${"2**31"}
+    ${Number.MAX_SAFE_INTEGER}     | ${BigInt(Number.MAX_SAFE_INTEGER)} | ${"bigint"}  | ${"@long"}   | ${"2**53 - 1"}
+    ${Number.MAX_SAFE_INTEGER + 1} | ${Number.MAX_SAFE_INTEGER + 1}     | ${"number"}  | ${"@double"} | ${"2**53"}
```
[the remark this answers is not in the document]
ptpaterson: `Number.MAX_SAFE_INTEGER + 1` is beyond the safe integer range for JS numbers, so at that point you have lost precision and shouldn't try to save that as a long.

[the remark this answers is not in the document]
cleve-fauna: see comment in source code for further explanation of where this came from.

__tests__/functional/tagged-format.test.ts
```diff
@@ -161,22 +164,49 @@ describe("tagged format", () => {
...
-    ${-(2 ** 63)}     | ${BigInt(-(2 ** 63))}     | ${"bigint"}  | ${"-2**63"}
-    ${2 ** 31 - 1}    | ${2 ** 31 - 1}            | ${"number"}  | ${"2**31 - 1"}
-    ${2 ** 31}        | ${BigInt(2 ** 31)}        | ${"bigint"}  | ${"2**31"}
-    ${2 ** 63 - 1}    | ${BigInt(2 ** 63 - 1)}    | ${"bigint"}  | ${"2**63 - 1"}
-    ${1.3 ** 63}      | ${1.3 ** 63}              | ${"number"}  | ${"1.3**63"}
-    ${1.3}            | ${1.3}                    | ${"number"}  | ${"1.3"}
+    input                          | expected                           | expectedType | tag          | testCase
+    ${LONG_MIN}                    | ${LONG_MIN}                        | ${"bigint"}  | ${"@long"}   | ${"-(2**63)"}
+    ${Number.MIN_SAFE_INTEGER - 1} | ${Number.MIN_SAFE_INTEGER - 1}     | ${"number"}  | ${"@double"} | ${"-(2**53)"}
+    ${Number.MIN_SAFE_INTEGER}     | ${BigInt(Number.MIN_SAFE_INTEGER)} | ${"bigint"}  | ${"@long"}   | ${"-(2**53 - 1)"}
```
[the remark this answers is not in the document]
ptpaterson: [MDN says](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger#description)

> The safe integers consist of all integers from -(2^53 - 1) to 2^53 - 1,

So the lowest number we can safely convert to Long is `-(2^53 - 1)`. Then the first whole number that can't be encoded as Long is `-(2^53 - 1) - 1 =>  -(2^53)`

__tests__/functional/tagged-format.test.ts
```diff
@@ -161,22 +164,49 @@ describe("tagged format", () => {
...
+    ${-(2 ** 31) - 1}              | ${BigInt(-(2 ** 31) - 1)}          | ${"bigint"}  | ${"@long"}   | ${"-(2**31) - 1"}
+    ${-(2 ** 31)}                  | ${-(2 ** 31)}                      | ${"number"}  | ${"@int"}    | ${"-(2**31)"}
+    ${0}                           | ${0}                               | ${"number"}  | ${"@int"}    | ${"0 (Int)"}
+    ${1}                           | ${1}                               | ${"number"}  | ${"@int"}    | ${"1 (Int)"}
+    ${BigInt("0")}                 | ${BigInt("0")}                     | ${"bigint"}  | ${"@long"}   | ${"0 (Long)"}
+    ${2 ** 31 - 1}                 | ${2 ** 31 - 1}                     | ${"number"}  | ${"@int"}    | ${"2**31 - 1"}
+    ${2 ** 31}                     | ${BigInt(2 ** 31)}                 | ${"bigint"}  | ${"@long"}   | ${"2**31"}
+    ${Number.MAX_SAFE_INTEGER}     | ${BigInt(Number.MAX_SAFE_INTEGER)} | ${"bigint"}  | ${"@long"}   | ${"2**53 - 1"}
+    ${Number.MAX_SAFE_INTEGER + 1} | ${Number.MAX_SAFE_INTEGER + 1}     | ${"number"}  | ${"@double"} | ${"2**53"}
+    ${LONG_MAX}                    | ${LONG_MAX}                        | ${"bigint"}  | ${"@long"}   | ${"2**64 - 1"}
```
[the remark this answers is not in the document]
ptpaterson: LONG_MAX and LONG_MIN are `BigInt`. `BigInt` values outside of that range throw errors. All plain JS numbers that fail [`Number.isSafeInteger()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger) are encoded as double. So testing if a Number is bigger than MAX_LONG is not necessary.

src/client-configuration.ts
```diff
@@ -6,6 +8,11 @@ export interface ClientConfiguration {
    * The {@link URL} of Fauna to call. See {@link endpoints} for some default options.
    */
   endpoint: URL;
+  /**
+   * Determines the encoded format expected for the query `arguments` field, and
+   * the `data` field of a successful response.
+   */
+  format: ValueFormat;
```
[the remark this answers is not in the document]
ptpaterson: The default value provided allows the user to not have to define it. But since it is safe to leave out the header, then we can do just that. The additional documentation is also very nice.

src/tagged-type.ts
ptpaterson: Fauna accepts the string representation with the exponent.

> mutate places we lose precision to doubles; and only blow up if we go past Fauna or JS's max double size.

I believe that we do this and that. The encoder checks for `Number.isSafeInteger`. If it is, then it's a Long. If not, then we encode as a Double.

cleve-fauna: we don't do that for BigInts.

src/tagged-type.ts
cleve-fauna: we will lose precision here too. We need to make a contract on what is acceptable to us.

example:

```
> 2**100
1.2676506002282294e+30
> 2**100 - 1
1.2676506002282294e+30
>
```

This is what i'm trying to get at with precision loss being treated the same for BigInt and number (e.g. do we allow it or blow up?)

src/tagged-type.ts
cleve-fauna: example:

```
> 2**100
1.2676506002282294e+30
> 2**100 - 1
1.2676506002282294e+30
>
```

This is what i'm trying to get at with precision loss being treated the same for BigInt and number (e.g. do we allow it or blow up?)

ptpaterson: No, but I think that is a good idea. Once you have a BigInt, the intent is clear that it should remain a Long. You can't even do arithmatic with a BigInt without converting other values to BigInt (e.g. `LONG_MAX + BigInt(1))` just to increment). BigInt is just a stand in for a 64bit Long. It seems better to me to throw an error that the int is too big before we convert to a Double on behalf of devs.

__tests__/functional/tagged-format.test.ts
```diff
@@ -161,22 +164,49 @@ describe("tagged format", () => {
...
       expect(typeof result.data).toEqual(expectedType);
-      expect(result.data).toEqual(expected);
+      expect(result.data.toString()).toEqual(expected.toString());
     }
   );
+
+  it.each`
+    input                   | testCase
+    ${LONG_MIN - BigInt(1)} | ${"lower than -(2**63) - 1"}
+    ${LONG_MAX + BigInt(1)} | ${"greater than 2**63"}
```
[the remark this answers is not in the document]
ptpaterson: Per our conversation, we will move forward with throwing an error when BigInt is outside of min/max Long values.

I added tests to throw if `Number.POSITIVE_INFINITY` or `Number.NEGATIVE_INFINITY` are encountered.

src/client-configuration.ts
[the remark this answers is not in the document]
ptpaterson: Done

src/client.ts
```diff
@@ -231,20 +236,23 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+        "x-format": this.#clientConfiguration.format,
       };
       this.#setHeaders(
         { ...this.clientConfiguration, ...queryRequest },
         headers
       );
 
+      const isTaggedFormat =
+        this.#clientConfiguration.format === "tagged" ||
+        queryRequest.format === "tagged";
```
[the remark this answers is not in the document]
ptpaterson: I updated this. The logic is the same, but I still referenced the client and request configurations. It could be slightly more terse if we used the defined header value, but then we'd have to get `headers["x-format"]`, which is not typed (and also a bit uglier IMO).

src/client.ts
```diff
@@ -26,10 +26,15 @@ import {
...
 } from "./http-client";
+import { TaggedTypeFormat } from "./tagged-type";
 
-const defaultClientConfiguration = {
-  max_conns: 10,
+const defaultClientConfiguration: Pick<
+  ClientConfiguration,
+  "format" | "endpoint" | "max_conns" | "timeout_ms"
+> = {
+  format: "tagged",
```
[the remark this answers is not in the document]
ptpaterson: Done

__tests__/functional/tagged-format.test.ts
```diff
@@ -161,22 +164,51 @@ describe("tagged format", () => {
...
+    ${LONG_MAX + BigInt(1)}     | ${"greater than 2**63"}
+    ${Number.NEGATIVE_INFINITY} | ${"NEGATIVE_INFINITY"}
+    ${Number.POSITIVE_INFINITY} | ${"POSITIVE_INFINITY"}
+  `("Throws if BigInt value is $testCase", async ({ input }) => {
+    expect.assertions(2);
+    try {
+      const result = await client.query(fql`${input}`);
+      console.log(result);
+    } catch (e) {
+      if (e instanceof ClientError) {
```
cleve-fauna, line 208: This should not be a client error - a client error implies a bug in our code.

Let's get some way to throw out an error here that clearly indicates to the user they sent us bad data.

(Perhaps we need to define a a basic IllegalArgumentError in the driver).

Let's cut a card for this for now.

src/wire-protocol.ts
```diff
@@ -47,6 +52,8 @@ export interface QueryRequestHeaders {
   traceparent?: string;
 }
 
+export type ValueFormat = "simple" | "tagged"
```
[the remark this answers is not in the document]
ptpaterson: Just committed this one.

src/client.ts
```diff
@@ -231,20 +235,22 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
-        // include per-request headers
       };
       this.#setHeaders(
         { ...this.clientConfiguration, ...queryRequest },
         headers
       );
 
+      const isTaggedFormat =
+        (this.#clientConfiguration.format ?? "tagged") === "tagged" ||
+        queryRequest.format === "tagged";
```
cleve-fauna, line 246: why not just use the derived headers? This is fine though.

cleve-fauna (approved):
Ship it. Please cut the card requested in the comment.

__tests__/functional/tagged-format.test.ts
cleve-fauna: Using: https://faunadb.atlassian.net/browse/FE-2990

src/client.ts
ptpaterson: This way it stays typed, but I don't have a strong argument outside of that. I like this better, although it is more verbose.

cleve-fauna: all good. that makes sense to me.

PR #72, opened 2023-02-27
Provide common module to get a test client

ptpaterson:
## Problem
Some of the tests hard-code using the docker images that don't need to. Some tests use `FAUNA_SECRET` as the env secret and others use `secret`.  I cannot update the wire protocol to what is in dev because of the tests tied to the docker image. 

## Solution
Every test that can should use the `FAUNA_SECRET` and `FAUNA_ENDPOINT` for consistency within the driver and conventions typical when using the driver.

Tests that must be run against the local image should still do that.

The env should not be polluted by any individual test.

## Result
I created a module that exports a `getClient` function. It either uses the environment vars or falls back on defaults for the local image. It takes the same arguments as the Client constructor.

## Testing
tests still pass when using local. Tests failing in other environments are due to wire protocol changes there.


lib/wire-protocol.d.ts
```diff
@@ -50,6 +50,7 @@ export interface QueryRequestHeaders {
      */
     traceparent?: string;
 }
+/** tagged declares that type information is transmitted and received by the driver. "simple" indicates it is not. */
```
cleve-fauna, line 53: i'm confused about why this just came in now.

ptpaterson: oh!  I bet becayse the last commit of the previous PR was done using the `commit suggestion` button

cleve-fauna (approved):

lib/wire-protocol.d.ts
cleve-fauna: oh yeah good call.

cleve-fauna: we need to clean up the build process to get it out of the commit hook 😵‍💫

PR #76, opened 2023-03-01
Refactor QueryBuilder to use composed wire format

ptpaterson:
Ticket(s): [FE-3091](https://faunadb.atlassian.net/browse/FE-3091)

## Problem
The database can now natively handle interpolated query fragments. We currently do a lot of extra work in the driver to glue a query together.

## Solution
Update the QueryBuilder implementation used with the `fql` tagged template function.

## Result
- Added types and documentation to the wire-protocol module.
- value fragments must _always_ be tagged/encoded, so 
  - require the fragments to be of type `TaggedTypeEncoded`.
  - do the encoding in the QueryBuilder implementation

## Out of scope
The encoder can throw `TypeError`s when precision would be lost (BigInt is too big). `QueryBuilder.toQuery()` is called higher up the chain (in `Client.query`) than encoding other arguments (in `Client.#query`), where encoding arguments is optional. We still need to determine how to unify and rethrow `TypeErrors` which are not "Client errors" per se, so I did not try to catch and rethrow any errors in `Client.query`.

## Testing
Updated the existing unit tests and added integration tests to send the composed wire format to the database.

[FE-3091]: https://faunadb.atlassian.net/browse/FE-3091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


__tests__/functional/tagged-format.test.ts
```diff
@@ -198,7 +198,10 @@ describe("tagged format", () => {
   `("Throws if BigInt value is $testCase", async ({ input }) => {
     expect.assertions(2);
     try {
-      const result = await client.query(fql`${input}`);
+      const result = await client.query({
+        query: "foo",
+        arguments: { foo: input },
+      });
```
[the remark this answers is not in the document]
ptpaterson: using the template format meant the encoding would happen inside of `TemplateQueryBuilder.toQuery` which broke the test since it's not rethrown as a ClientError. I didn't want to change what we were doing too much.

lib/tagged-type.d.ts
```diff
@@ -23,3 +23,8 @@ export declare class TaggedTypeFormat {
 }
 export declare const LONG_MIN: bigint;
 export declare const LONG_MAX: bigint;
+export declare class TaggedTypeEncoded {
+    #private;
```
cleve-fauna, line 27: what does this do? Can't find docs on this here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields

__tests__/functional/tagged-format.test.ts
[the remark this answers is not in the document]
cleve-fauna: i see thanks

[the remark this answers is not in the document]
ptpaterson: alternative is to do a pure check that
```javascript
expects(client.query(fql`input`)).throws()
```

lib/tagged-type.d.ts
ptpaterson: I dunno! that's some weird typescript magic happening right there in the build files.

cleve-fauna: oh lol that's the build file 😵‍💫 

got to move that!

src/wire-protocol.ts
```diff
@@ -1,11 +1,60 @@
+import { fql } from "./query-builder";
+import { TaggedTypeEncoded } from "./tagged-type";
+
+/**
+ * A piece of an interpolated query that represents an actual value. Arguments
+ * are passed to fauna using ValuFragments so that query string injection is not
```
cleve-fauna, line 6:
```suggestion
 * are passed to fauna using ValueFragments so that query string injection is not
```

src/wire-protocol.ts
```diff
@@ -1,11 +1,60 @@
...
+ *  const num = 17;
+ *  const query = fql`${num} + 3)`;
+ *  // produces
+ *  { fql: [{ value: { "@int": "17" } }, " + 3"] }
+ * ```
+ */
+export type ValueFragment = { value: TaggedTypeEncoded };
+
+/**
+ * A piece of an interpolated query. Queries can be safely composed together
```
cleve-fauna, line 25:
```suggestion
 * A piece of an interpolated query. Interpolated queries can be safely composed together
```

src/wire-protocol.ts
```diff
@@ -1,11 +1,60 @@
...
+ *  const query1 = fql`${num} + 3)`;
+ *  const query2 = fql`5 + ${query1})`;
+ *  // produces
+ *  { fql: ["5 + ", { fql: [{ value: { "@int": "17" } }, " + 3"] }] }
+ * ```
+ */
+export type FQLFragment = { fql: (string | QueryInterpolation)[] };
+
+/**
+ * A piece of an interpolated query. Queries can be safely composed together
```
cleve-fauna, line 41:
```suggestion
 * A piece of an interpolated query. Interpolated queries can be safely composed together
```

src/wire-protocol.ts
ptpaterson: I like it 👍🏻

cleve-fauna: we'll have to regen lib once suggestions are merged 😵‍💫

src/query-builder.ts
```diff
@@ -23,157 +26,92 @@ export interface QueryBuilder {
...
+        const arg = this.#queryArgs[i];
+        let subQuery: string | QueryInterpolation;
+        if (isQueryBuilder(arg)) {
+          const request = arg.toQuery(requestHeaders);
+          subQuery = request.query;
+          resultArgs = { ...resultArgs, ...request.arguments };
+        } else {
+          // arguments in the template format must always be encoded, regardless
+          // of the "x-format" request header
+          // TODO: catch and rethrow TypeErrors as Fauna errors
```
cleve-fauna, line 105:
```suggestion
          // TODO: catch and rethrow TypeErrors as Fauna errors indicating bad user input
```

src/wire-protocol.ts
```diff
@@ -1,11 +1,60 @@
...
+ * request header sent.
+ * @example
+ * ```typescript
+ *  const num = 17;
+ *  const query = fql`${num} + 3)`;
+ *  // produces
+ *  { fql: [{ value: { "@int": "17" } }, " + 3"] }
+ * ```
+ */
+export type ValueFragment = { value: TaggedTypeEncoded };
```
cleve-fauna, line 22: Nit:

Thinking of readers here trying to build their own drivers - it will be easier to read if the "public" types appear first and the "private" types appear last:

- Please move these type definitions to the bottom of the file; or at least below the request / response models. (Alternatively, put them in another file). Do what you think works best.

cleve-fauna (commented):
Looks very good!

Some suggestions on documentation and additional tests.

__tests__/integration/template-format.test.ts
```diff
@@ -0,0 +1,100 @@
+import { fql } from "../../src/query-builder";
+import { getClient } from "../client";
+
+const client = getClient({
+  max_conns: 5,
+  query_timeout_ms: 60_000,
+});
+
+describe("query using template format", () => {
+  it("succeeds with a string variable", async () => {
```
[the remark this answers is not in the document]
ptpaterson: > Can you also add a test case that leverages the QueryBuilder + server side string interpolation
You mean use `fql()` and provide a small QueryBuilder implementation as an arg?

Something like this?
```javascript
const innerBuilder = {
  toQuery: () => { query: "arg0", arguments: { arg0: "hello" }
}
const queryBuilder = fql`"${innerBuilder} world"`
// ...
```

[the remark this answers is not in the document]
cleve-fauna: no.

something like this:

```javascript
let ageTwoYearsAgo = 3;
let twoYears = 2;
fql`Alice is #{${ageTwoYearsAgo} + ${twoYears}} old`;
```

Like: https://deploy-preview-1272--fauna-docs.netlify.app/fqlx/beta/reference/language/literals#interpolated-strings

__tests__/unit/query-builder.test.ts
```diff
@@ -4,89 +4,91 @@ describe("fql method producing QueryBuilders", () => {
   it("parses with no variables", () => {
```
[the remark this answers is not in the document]
ptpaterson: The general case is handled here: 

https://github.com/fauna/fauna-js/blob/32531278170a488bd77e89bc8fff1f924bcbfc8d/__tests__/integration/query.test.ts#L63-L66

Is there something more you are looking for?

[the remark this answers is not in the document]
ptpaterson: nvm I misunderstood "server side string interpolation" but I got it now 👍🏻

__tests__/integration/template-format.test.ts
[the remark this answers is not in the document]
ptpaterson: ah got it! I also like the "works with other query builders" test, too. So I'll do that too.

src/wire-protocol.ts
ptpaterson: done

src/wire-protocol.ts
ptpaterson: done (and new build too 😆)

src/wire-protocol.ts
ptpaterson: done

src/query-builder.ts
ptpaterson: done

src/wire-protocol.ts
ptpaterson: I dropped them down just above errors. I think I'd like to get errors moved into it's own module too, part of some reorganizing and dealing with properly exporting stuff to users.

__tests__/integration/template-format.test.ts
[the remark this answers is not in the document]
ptpaterson: test added

__tests__/unit/query-builder.test.ts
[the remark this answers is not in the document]
ptpaterson: test added

cleve-fauna (approved):

PR #78, opened 2023-03-02
Throw RangeError, not TypeError, when encoding out-of-range numbers

ptpaterson:
Ticket(s): [FE-2990](https://faunadb.atlassian.net/browse/FE-2990)

## Problem
We throw a `TypeError` in a couple of places, but these should be `RangeError` instead. According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError): 

> The RangeError object indicates an error when a value is not in the set or range of allowed values.

One of the [examples of `RangeError` includes throwing when infinite values cannot be handled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Resulting_string_too_large).

## Solution
Swap out the error types and test that relies on it.

[FE-2990]: https://faunadb.atlassian.net/browse/FE-2990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lib/wire-protocol.d.ts
```diff
@@ -110,7 +110,6 @@ export declare const isQuerySuccess: (res: any) => res is QuerySuccess<any>;
 export declare const isQueryFailure: (res: any) => res is QueryFailure;
 export declare const isQueryResponse: (res: any) => res is QueryResponse<any>;
 /**
-<<<<<<< HEAD
```
pnwpedro, line 113: :D

src/wire-protocol.ts
```diff
@@ -132,7 +132,6 @@ export const isQueryResponse = (res: any): res is QueryResponse<any> =>
   isQueryResponse(res) || isQueryFailure(res);
 
 /**
-<<<<<<< HEAD
```
pnwpedro, line 135: :D

pnwpedro (approved):

src/wire-protocol.ts
ptpaterson: yea, snuck that in XD. I don't when it showed up, it caused tests to fail 🤷🏻

PR #79, opened 2023-03-02
Ignore the lib folder

ptpaterson:
Ticket(s): [FE-3108](https://faunadb.atlassian.net/browse/FE-3108)

This is step 1 of moving the build process to a pipeline.

## Problem
We don't need to commit the build files to the repo.

## Solution
Add `lib/` to `.gitignore` and also ignore the build folder in the pre-commit hook. We can still run the build as a pre-commit hook, ensuring that we don't commit TS compile errors.

## Out of scope
anything else regarding publishing the build

## Testing
I tested out the precommit hook with some bad TS to make sure that still worked without trying to commit the build folder.

[FE-3108]: https://faunadb.atlassian.net/browse/FE-3108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cleve-fauna:
I'm assuming this is step 1 of building a release pipeline?

PR #83, opened 2023-03-03
Organize request and response types

ptpaterson:
Ticket(s): [FE-3105](https://faunadb.atlassian.net/browse/FE-3105)

This PR shuffles types about how I have come to view them. I hope that it can be a good step toward clarifying the mental model for users of the driver.

## Problem
- Request and response types can be cleanly separated into different modules
- `QueryRequest` extends `QueryRequestHeaders`, but `QueryRequestHeaders` fields are not actually sent over the wire, which muddies the types.

## Solution
Separate `QueryRequest` and `QueryRequestHeaders` so we can enforce that only a QueryRequest provided to the HTTPClient and thus over the wire. Move `QueryRequestHeaders` out of the wire-protocol module.

### Separate types into multiple files
Created separate files for request and response types. They are still in a folder called `wire-protocol' so imports don't actually change.

### Query Options
`QueryRequestHeaders` was also renamed to `QueryRequestOptions` to reflect that they could be any options (though they happen to currently only include specific header values).

Instead of duplicate the same fields in `ClientConfiguration`, moved those into an embedded, optional `queryOptions` field that is a `QueryRequestOptions` object. I documented that these are the options sent be default for each query. This way, we have a consistent story about what these options are: they are defined in a single place, and you can provide some of the options as client defaults, and you can also provide any of the options per query.

Embedding the query options breaks the existing API. It makes client configuration (when you need to set query options) a little more verbose, but I think makes it more clear what you are doing.

Also, this separation also makes type enforcement work better. Before, when client figuration was essentially `ClientOptions & QueryOptions` but there are some places (like `Client.#setHeaders`) that you only want the query options.

### JSON types

Also noted that `JSONObject` and `JSONValue` are necessary because we are not handling encoding/decoding of user classes.

[FE-3105]: https://faunadb.atlassian.net/browse/FE-3105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -241,13 +241,13 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
         "x-typecheck": "false",
       };
       this.#setHeaders(
-        { ...this.clientConfiguration, ...queryRequest },
+        { ...this.clientConfiguration.queryOptions, ...queryRequest },
         headers
       );
 
       const isTaggedFormat =
-        (this.#clientConfiguration.format ?? "tagged") === "tagged" ||
-        queryRequest.format === "tagged";
+        (this.#clientConfiguration.queryOptions?.format ?? "tagged") ===
+          "tagged" || queryRequest.format === "tagged";
       const queryArgs = isTaggedFormat
         ? TaggedTypeFormat.encode(queryRequest.arguments)
         : queryRequest.arguments;
--- a/src/client.ts
+++ b/src/client.ts
@@ -1,4 +1,8 @@
-import { ClientConfiguration, endpoints } from "./client-configuration";
+import {
+  ClientConfiguration,
+  endpoints,
+  QueryRequestOptions,
+} from "./client-configuration";
 import {
   AuthenticationError,
   AuthorizationError,
@@ -19,7 +23,6 @@ import {
   isQuerySuccess,
   type QueryFailure,
   type QueryRequest,
-  type QueryRequestHeaders,
   type QuerySuccess,
 } from "./wire-protocol";
 import {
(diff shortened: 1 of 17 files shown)


src/client.ts
```diff
@@ -138,13 +143,13 @@ export class Client {
    * due to an internal error.
    */
   async query<T = any>(
-    request: QueryRequest | QueryBuilder,
-    headers?: QueryRequestHeaders
+    request: (QueryRequest & QueryRequestOptions) | QueryBuilder,
+    requestOptions?: QueryRequestOptions
   ): Promise<QuerySuccess<T>> {
```
ptpaterson, line 148: Two places this change turned `QueryRequest` into `(QueryRequest & QueryRequestOptions)`. Which is kinda ugly.

> Note, you can embed request options in this object; if you do that there's no need to pass the requestOptions parameter.

I would like to argue that we shouldn't overload the request parameter like this. But right now this PR doesn't change any functionality, whereas dropping `& QueryRequestOptions` would break the contract we've had for a long time.

cleve-fauna: @ptpaterson its part of the wire protocol that any header can be sent as a request field.

That's why it was done this way.

cleve-fauna:
@ptpaterson IMO this makes things a bit confusing.

The `QueryRequestHeaders` type showed users what the valid headers were.

And the wire protocol allows for every possible header to also be sent in the request body which was the reason for the original implementation.

`QueryRequestOptions` just seems like an oblique name for headers.

src/client.ts
ptpaterson: well hell

ptpaterson: Actually... not anymore? I don't see that in the openAPI spec.

ptpaterson:
That makes sense, but then I think the spec must have changed before I jumped on board.

I'll need to try this out. And also dig into the spec history to see when this changed, and be really sure it has changed.

src/client.ts
ptpaterson: Confirmed by core team that we've removed the options from the query body and only accept through headers.

PR #82, opened 2023-03-03
Remove max_conns configuration

ptpaterson:
Ticket(s): [FE-3290](https://faunadb.atlassian.net/browse/FE-3290)

## Problem
After removing axios, the `max_conns` configuration is not necessary to support the fetch or http2 APIs.

## Solution
Remove `max_conns` as a client configuration options and from the Client, tests, demo, and READMEs

There was no place where the max_conns configuration was being used, so no substantive changes were made.

[FE-3109]: https://faunadb.atlassian.net/browse/FE-3109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FE-3290]: https://faunadb.atlassian.net/browse/FE-3290?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #85, opened 2023-03-06
Node http2

ptpaterson:
Ticket(s): [FE-3061](https://faunadb.atlassian.net/browse/FE-3061)

## Problem
We want to start using the native http2 module when we detect we're running in the Node env.

## Solution
Add an HTTPClient implementation that uses the http2 Node module. Client should use it as default when in Node.

## Result
- [x] Provide HTTPClient implementation using the Node built in `http2` package
- [x] Browser build excludes the http2 package
- [x] Implemented default session idle time at 500ms to prevent 
- [x] NodeHTTP2Client class is flexible enough for streams, including handling the URL path and passing stream-consumer options.

## Testing
Added test to confirm the new NodeHTTP2Client is used by default. That means that all of the tests that do not explicitly provide a FetchClient are now using NodeHTTP2Client.

[FE-3061]: https://faunadb.atlassian.net/browse/FE-3061?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/http-client/index.ts
```diff
@@ -38,13 +40,15 @@ export interface HTTPClient {
...
 
 // utility functions
 
 export const isHTTPResponse = (res: any): res is HTTPResponse =>
   res instanceof Object && "body" in res && "headers" in res && "status" in res;
+
+const isNode = () =>
+  Object.prototype.toString.call(
+    typeof process !== "undefined" ? process : 0
+  ) === "[object process]";
```
cleve-fauna, line 54: I haven't looked at this whole PR yet - so more may come later! But I think we can be a bit tighter here:

```suggestion
const isNode = () =>  (process !== undefined && process.release === "node");
```

Example:

```
 > process.release
{ name: 'node', lt
  name: 'node',
  lts: 'Hydrogen',
  sourceUrl: 'https://nodejs.org/download/release/v18.12.1/node-v18.12.1.tar.gz',
  headersUrl: 'https://nodejs.org/download/release/v18.12.1/node-v18.12.1-headers.tar.gz'
}
 > 
```

src/http-client/node-http2-client.ts
```diff
@@ -0,0 +1,222 @@
...
+      return this.#client;
+    }
+
+    this.#client = new NodeHTTP2Client();
+    return this.#client;
+  }
+
+  /** {@inheritDoc HTTPClient.request} */
+  async request(httpRequest: HTTPRequest): Promise<HTTPResponse> {
+    const session = this.#getSession(httpRequest.url);
```
cleve-fauna, line 35: Let's take a P2 or P3 to have the maxConns of the ClientConfiguration control how many streams we open per session: https://nodejs.org/api/http2.html#collecting-http2-performance-metrics

There's an argument there that some clients may want to use that as a throttle on how much work their client is doing.

#82 fits well with it because we should rename max conns to be something like "max streams"

Let's sync with the go and python driver crew to see what they are doing here as well. Not blocking for P1

cleve-fauna (commented):
Looks pretty good.

I think we can simplify the implmentation by planning on supporting our "streaming feature" with a dedicated HTTPClient.

That would allow us to use Node's native session timeout facilities in this "request/reply" client instead of rolling our own.

Curious to hear your thoughts on that.

src/http-client/node-http2-client.ts
ptpaterson: I created [FE-3290](https://faunadb.atlassian.net/browse/FE-3290) to track outside of P1 fit and finish

src/http-client/node-http2-client.ts
```diff
@@ -0,0 +1,222 @@
...
+
+    try {
+      const result = await session.request(httpRequest);
+      return result;
+    } catch (error) {
+      // We do this cleanup in v4 driver. Does it risk killing other open requests
+      // that we want to let go, or if we are in this catch block are we actually
+      // safe to do this?
+      this.#cleanupSession(httpRequest.url);
```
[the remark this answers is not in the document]
ptpaterson: Removed the cleanup call here. Other calls to `#cleanupSession` I changed to just `session.close()` to keep things simple.

We'll need to get some real use out of this or figure out some tests to exercise these error paths; I am not sure what, if any, errors are going to get silently swallowed the way things are now.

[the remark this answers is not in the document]
ptpaterson: And that's a separate comment from the suggested change. Regardless, I am just looking back at the error handling here.

src/http-client/node-http2-client.ts
```diff
@@ -0,0 +1,222 @@
...
+      this.#setInactivityTimeout();
+    }
+  }
+
+  #onRequestStart() {
+    ++this.#ongoingRequests;
+    this.#clearInactivityTimeout();
+  }
+
+  #setInactivityTimeout() {
```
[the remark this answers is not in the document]
ptpaterson: Just pushed this change.

src/http-client/node-http2-client.ts
```diff
@@ -0,0 +1,169 @@
...
+import { NetworkError } from "../errors";
+import { QueryRequest } from "../wire-protocol";
+
+const { HTTP2_HEADER_PATH, HTTP2_HEADER_METHOD, HTTP2_HEADER_STATUS } =
+  http2.constants;
+
+/**
+ * An implementation for {@link HTTPClient} that uses the node http package
+ */
+export class NodeHTTP2Client implements HTTPClient {
```
cleve-fauna, line 13: P2 - expose a close method in the HTTPClient interface that allows requests to complete and then closes down the session.

cleve-fauna (approved):
🐑

src/http-client/node-http2-client.ts
cleve-fauna: i'll cut a JIRA for this.

PR #87, opened 2023-03-08
Date time wrappers

ptpaterson:
Ticket(s): [FE-3080](https://faunadb.atlassian.net/browse/FE-3080)

## Problem
Fauna's `Time` type can store nanosecond precision. Javascript `Date` can only store milliseconds.

There is no built in type for "plain" dates in JS. Closest thing is using a `Date` with time set to 00:00:00 UTC. So Javascript `Date` is not well suited to store Fauna `Date` values.

## Solution
Provide our own type, very simple wrappers a la v4 driver. Stores the exact encoded string value. 

Native Javascript `Date`s are always encoded as a Fauna `Time`.

Users may still need JS `Date`s to do date/time things with it, so they may lose precision when converting to and from.

## Testing
Added a unit tests for constructing and deconstructing FaunaTime and FaunaDate. Updated tests for encoding and decoding.

[FE-3080]: https://faunadb.atlassian.net/browse/FE-3080?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/tagged-type.ts
+++ b/src/tagged-type.ts
@@ -132,8 +132,8 @@ export class TaggedTypeEncoded {
     date: (dateValue: Date): TaggedTime => ({
       "@time": dateValue.toISOString(),
     }),
-    faunadate: (value: DateStub): TaggedDate => ({ "@date": value.value }),
-    faunatime: (value: TimeStub): TaggedTime => ({ "@time": value.value }),
+    faunadate: (value: DateStub): TaggedDate => ({ "@date": value.dateString }),
+    faunatime: (value: TimeStub): TaggedTime => ({ "@time": value.isoString }),
   };
 
   constructor(input: any) {
--- a/src/tagged-type.ts
+++ b/src/tagged-type.ts
@@ -1,4 +1,4 @@
-import { FaunaDate, FaunaTime } from "./values";
+import { DateStub, TimeStub } from "./values";
 
 /** A reference to a built in Fauna module; e.g. Date */
 export type Module = string;
@@ -51,9 +51,9 @@ export class TaggedTypeFormat {
       } else if (value["@double"]) {
         return Number(value["@double"]);
       } else if (value["@date"]) {
-        return FaunaDate.from(value["@date"]);
+        return DateStub.from(value["@date"]);
       } else if (value["@time"]) {
-        return FaunaTime.from(value["@time"]);
+        return TimeStub.from(value["@time"]);
       } else if (value["@object"]) {
         return value["@object"];
       }
@@ -132,8 +132,8 @@ export class TaggedTypeEncoded {
     date: (dateValue: Date): TaggedTime => ({
       "@time": dateValue.toISOString(),
     }),
-    faunadate: (value: FaunaDate): TaggedDate => ({ "@date": value.value }),
-    faunatime: (value: FaunaTime): TaggedTime => ({ "@time": value.value }),
+    faunadate: (value: DateStub): TaggedDate => ({ "@date": value.value }),
+    faunatime: (value: TimeStub): TaggedTime => ({ "@time": value.value }),
   };
 
   constructor(input: any) {
@@ -157,9 +157,9 @@ export class TaggedTypeEncoded {
           this.result = this.#encodeMap["array"](input);
         } else if (input instanceof Date) {
           this.result = this.#encodeMap["date"](input);
-        } else if (input instanceof FaunaDate) {
+        } else if (input instanceof DateStub) {
           this.result = this.#encodeMap["faunadate"](input);
-        } else if (input instanceof FaunaTime) {
+        } else if (input instanceof TimeStub) {
           this.result = this.#encodeMap["faunatime"](input);
         } else {
           this.result = this.#encodeMap["object"](input);
--- a/src/tagged-type.ts
+++ b/src/tagged-type.ts
@@ -1,3 +1,5 @@
+import { FaunaDate, FaunaTime } from "./values";
+
 /** A reference to a built in Fauna module; e.g. Date */
 export type Module = string;
 /** A reference to a document in Fauna */
@@ -49,9 +51,9 @@ export class TaggedTypeFormat {
       } else if (value["@double"]) {
         return Number(value["@double"]);
       } else if (value["@date"]) {
-        return new Date(value["@date"] + "T00:00:00.000Z");
+        return FaunaDate.from(value["@date"]);
       } else if (value["@time"]) {
-        return new Date(value["@time"]);
+        return FaunaTime.from(value["@time"]);
       } else if (value["@object"]) {
         return value["@object"];
       }
@@ -127,18 +129,11 @@ export class TaggedTypeEncoded {
       for (const i in input) _out.push(TaggedTypeFormat.encode(input[i]));
       return _out;
     },
-    date: (dateValue: Date): TaggedDate | TaggedTime => {
-      if (
-        dateValue.getUTCHours() == 0 &&
-        dateValue.getUTCMinutes() == 0 &&
-        dateValue.getUTCSeconds() == 0 &&
-        dateValue.getUTCMilliseconds() == 0
-      ) {
-        return { "@date": dateValue.toISOString().split("T")[0] };
-      }
-
-      return { "@time": dateValue.toISOString() };
-    },
+    date: (dateValue: Date): TaggedTime => ({
+      "@time": dateValue.toISOString(),
+    }),
+    faunadate: (value: FaunaDate): TaggedDate => ({ "@date": value.value }),
+    faunatime: (value: FaunaTime): TaggedTime => ({ "@time": value.value }),
   };
 
   constructor(input: any) {
@@ -162,6 +157,10 @@ export class TaggedTypeEncoded {
           this.result = this.#encodeMap["array"](input);
         } else if (input instanceof Date) {
           this.result = this.#encodeMap["date"](input);
+        } else if (input instanceof FaunaDate) {
+          this.result = this.#encodeMap["faunadate"](input);
+        } else if (input instanceof FaunaTime) {
+          this.result = this.#encodeMap["faunatime"](input);
         } else {
           this.result = this.#encodeMap["object"](input);
         }
--- a/src/values.ts
+++ b/src/values.ts
@@ -1,3 +1,6 @@
+import { parse } from "path";
+import { ClientError } from "./errors";
+import * as PARSE from "./regex";
 /**
  * An wrapper around the Fauna `Time` type. It, represents a fixed point in time
  * without regard to calendar or location, e.g. July 20, 1969, at 20:17 UTC.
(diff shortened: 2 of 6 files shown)


src/values.ts
```diff
@@ -1,21 +1,166 @@
...
+  static fromDate(date: Date): FaunaTime {
+    return new FaunaTime(date.toISOString());
+  }
+
+  /**
+   * Get the underlying string value
+   */
+  get value(): string {
+    return this.#value;
+  }
```
cleve-fauna, line 62: I would call this isoString

makes it clear to users what it is.

value is a bit ambiguous

src/values.ts
```diff
@@ -1,21 +1,166 @@
...
+ * This FaunaTime class preserves precision by storing the original string value
+ * and should be used whenever possible to pass `Time` values back to Fauna.
+ * Converting to a Javascript date before sending to Fauna could result in loss
+ * of precision.
+ *
+ * TODO: link to beta docs for `Time` type
  */
-class FaunaTime {}
+export class FaunaTime {
+  readonly #value: string;
```
cleve-fauna, line 16: i would just call this isoString

src/values.ts
```diff
@@ -1,21 +1,166 @@
...
+    return this.#value;
+  }
+
+  /**
+   * Get a copy of the `FaunaTime` converted to a Javascript `Date`. Does not
+   * mutate the existing `FaunaTime` value.
+   * @returns A `Date`
+   */
+  toDate(): Date {
+    return new Date(this.#value);
```
cleve-fauna, line 70: the contract on the Date constructor is weak - it only supports a subset of ISO-8601.

Looks like parse: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse has a better contract.

So we could do:

```
return new Date(Date.parse(this.#value));
```

src/values.ts
```diff
@@ -1,21 +1,166 @@
...
+ * timezone info by default.
  *
- * @remarks This class follows the TC39 proposal for `Temporal.PlainDate`. See the
- * proposal here: https://tc39.es/proposal-temporal/docs/index.html.
- * Since fauna has both `Time` and `Date` data types, we want to make a clear
- * distinction between them and avoid using the native `Date` class.
+ * TODO: link to beta docs for `Date` type
  */
-class FaunaDate {}
+export class FaunaDate {
```
cleve-fauna, line 95: same comments on naming and weakeness of Date constructor contract

cleve-fauna (changes requested):
naming changes and using different date apis would clean things up communication wise to users + correctness wise on conversions

cleve-fauna:
FWIW its we can link to docs here:

https://deploy-preview-1272--fauna-docs.netlify.app/fqlx/beta/reference/builtin_functions/time/time

these will be publicly available for beta. so we can slot those in for now

src/values.ts
ptpaterson: Something like `"01 Jan 1970 00:00:00 GMT"` passes `Date.parse` but Fauna cannot accept it.

cleve-fauna: yeah that's why i want to move off of it.

cleve-fauna: we should only accept ISO-8601 strings for clarity.

cleve-fauna: sorry i misread this. Let me read the MDN docs again.

cleve-fauna: Ah I see the docs say parse only _explicitly_ supports ISO-8601 and that other format behavior is browser specific.

Yikes.

Looks like Date constructor has similar issues - it will take non ISO-8601.

I'd propose the following:

1. Let's maintain an ISO-8601 regex.
2. Verify the input matches the regex
3. Use parse to avoid Date constructor's pitfalls

ptpaterson: That's a plan!

ptpaterson: intersect that java date-parsing problems 🙄 Fauna _requires_ the timezone part

image

ptpaterson: clarification, spec assumes that no TZ means "local" time and so up to you to handle. I figured we'd let Fauna assume UTC. We'll have more explicitly convert to UTC or bounce back to the user. An error would be the simplest thing, with a message that says a Time must be relative to a timezone.

cleve-fauna: sorry i don't quite follow your latest message here.

src/values.ts
```diff
@@ -1,21 +1,194 @@
...
+    if (!(date instanceof Date)) {
+      throw new TypeError(`Expected Date but received ${typeof date}: ${date}`);
+    }
+    const dateString = date.toISOString();
+    const matches = PARSE.startsWithPlaindate.exec(dateString);
+    if (matches === null) {
+      // Our regex should match any possible date that comes out of
+      // `Date.toISOString()`, so we will only get here if the regex is
+      // incorrect. This is a ClientError since it is our fault.
+      throw new ClientError(`Failed to parse date '${date}'`, { cause: null });
```
cleve-fauna, line 160:
```suggestion
      throw new ClientError(`Failed to parse date '${date}'`);
```

src/values.ts
```diff
@@ -1,21 +1,194 @@
...
+  /**
+   * Creates a new {@link DateStub} from a Javascript `Date`
+   * @param date - A Javascript `Date`. The time is converted to UTC before
+   * saving the date.
+   * @returns A new {@link DateStub}
+   */
+  static fromDate(date: Date): DateStub {
+    if (!(date instanceof Date)) {
+      throw new TypeError(`Expected Date but received ${typeof date}: ${date}`);
+    }
```
cleve-fauna, line 153: feels a bit non-standard to do type checks like this. we could drop them IMO. 

If customers want type checking let's point them to typescript and keep the type checking in JS minimal.

same comment elsewhere.

src/values.ts
```diff
@@ -1,21 +1,194 @@
...
+      throw new ClientError(`Failed to parse date '${date}'`, { cause: null });
+    }
+    return new DateStub(matches[0]);
+  }
+
+  /**
+   * Get the underlying string value
+   */
+  get dateString(): string {
+    return this.#dateString;
```
cleve-fauna, line 169: nit: you could just have the field be readonly and not be private

src/values.ts
```diff
@@ -1,21 +1,194 @@
...
+   */
+  static fromDate(date: Date): TimeStub {
+    return new TimeStub(date.toISOString());
+  }
+
+  /**
+   * Get the underlying string value
+   */
+  get isoString(): string {
+    return this.#isoString;
```
cleve-fauna, line 67: nit: you could just have this field be readonly and not private

src/values.ts
```diff
@@ -1,21 +1,194 @@
+import { parse } from "path";
+import { ClientError } from "./errors";
+import * as PARSE from "./regex";
 /**
- * An Instant represents a fixed point in time (called "exact time"), without
- * regard to calendar or location, e.g. July 20, 1969, at 20:17 UTC.
+ * An wrapper around the Fauna `Time` type. It, represents a fixed point in time
+ * without regard to calendar or location, e.g. July 20, 1969, at 20:17 UTC.
  *
```
cleve-fauna, line 7:
```suggestion
 * An wrapper around the Fauna `Time` type. It, represents a fixed point in time
 * without regard to calendar or location, e.g. July 20, 1969, at 20:17 UTC.
 * Convert to and from javascript Date's with the fromDate and toDate methods. See remarks for possible
 * precision loss when doing this. If precision loss is a concern consider using a 3rd party datetime library
 * such as luxon.
```

src/values.ts
```diff
@@ -1,21 +1,194 @@
...
+  toJSON(): string {
+    return this.#isoString;
+  }
+}
 
 /**
- * A Temporal.PlainDate object represents a calendar date that is not associated
- * with a particular time or time zone, e.g. August 24th, 2006.
+ * A wrapper aroud the Fauna `Date` type. It represents a calendar date that is
+ * not associated with a particular time or time zone, e.g. August 24th, 2006.
```
cleve-fauna, line 96:
```suggestion
 * A wrapper aroud the Fauna `Date` type. It represents a calendar date that is
 * not associated with a particular time or time zone, e.g. August 24th, 2006.
 * Convert to and from javascript Date's with the fromDate and toDate methods. See remarks for possible
 * the time information that will be added / stripped when doing this. If adding and stripping times is a concern
 * consider using a 3rd party library to represent your dates in app before converting to DateStub's for transmission
 * to Fauna. Example libraries include luxon.
```

cleve-fauna (commented):
Some suggestions

src/values.ts
```diff
@@ -1,21 +1,194 @@
...
+    return date;
+  }
+
+  /**
+   * Override default JSON output
+   * @returns the string representation of the Fauna Time
+   */
+  toJSON(): string {
+    return this.#dateString;
+  }
```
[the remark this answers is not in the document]
ptpaterson: I might like that better. What about you @cleve-fauna ? Current driver returns wire protocol when you `JSON.stringify` it and it can be handy.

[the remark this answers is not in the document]
cleve-fauna: I figured that was the use case; seems reasonable.

Let's clean up the documentation some. 

```
/**
 * @return JSON representation of this TimeStub - the ISO-8601 time string. 
*/
```
Similar for the DateStub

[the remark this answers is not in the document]
ptpaterson: oh but we have to worry about tagged vs untagged 😩 . I changed my mind, I think returning the ISO string is good.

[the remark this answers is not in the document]
cleve-fauna: I'd rather not implement it at all (until someone asks for it) or use the ISO 8601 string.

We definitely don't want users to build dependencies on the low level specifics of the wire protocol.

I can't think of a a reason to present the `@date` internal to the users. One of the values of teh DateStub is that it provides a clean wrapper / layer between the wire protocol and date libraries. so let's keep the internals of the wire protocol in the encoder / decoder.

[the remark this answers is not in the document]
cleve-fauna: Also, for FQL expressions emitting wire protocol - I like the approach @sprsquish took in Go and pushed Python to - that encoding / decoding is decoupled from composition.

If we refactored to that we would not spill the wire protocol there either.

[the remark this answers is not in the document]
cleve-fauna: but we can tackle that in a different body of work 😄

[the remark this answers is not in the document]
cleve-fauna: @ptpaterson i edited the big comment a couple comments up a bit.

src/values.ts
ptpaterson: Fauna requires all `Time` values to have a timezone part, e.g. `Z` or `+02:00` or `+2000`, etc.

If we _want_, when we construct a `TimeStub` directly from a string, we could catch that no timezone part was provided and assume UTC, i.e. append `"Z"` to the end.

The PR so far says if the string doesn't match then throw an error. It doesn't try anything more sophisticated. I think this is the right path. And we can always add support for parsing "local time" (i.e. ISO string with no TZ part) later.

src/values.ts
ptpaterson: Removing this causes a type error. We can make the `options` parameter options when constructing `ClientError`.

src/values.ts
ptpaterson: Ope! This fell out of the naming changes. Yea we can do that.

src/values.ts
cleve-fauna: ah okay yeah seems like it should be optional

src/values.ts
cleve-fauna: cool sounds right to me too

src/tagged-type.ts
```diff
@@ -127,18 +129,11 @@ export class TaggedTypeEncoded {
...
-      }
-
-      return { "@time": dateValue.toISOString() };
-    },
+    date: (dateValue: Date): TaggedTime => ({
+      "@time": dateValue.toISOString(),
+    }),
+    faunadate: (value: DateStub): TaggedDate => ({ "@date": value.dateString }),
+    faunatime: (value: TimeStub): TaggedTime => ({ "@time": value.isoString }),
   };
```
cleve-fauna, line 137: i'd call these `date_stub` and `time_stub` for clarity but OK to ship like this for now.

src/values.ts
```diff
@@ -1,21 +1,186 @@
+import { parse } from "path";
+import { ClientError } from "./errors";
+import * as PARSE from "./regex";
 /**
- * An Instant represents a fixed point in time (called "exact time"), without
- * regard to calendar or location, e.g. July 20, 1969, at 20:17 UTC.
+ * An wrapper around the Fauna `Time` type. It, represents a fixed point in time
```
cleve-fauna, line 5:
```suggestion
 * A wrapper around the Fauna `Time` type. It, represents a fixed point in time
```

OK to ship as is for now

cleve-fauna (approved):

src/values.ts
ptpaterson: Why do I keep doing that?!

src/tagged-type.ts
ptpaterson: oh good catch.

PR #89, opened 2023-03-10
Add classes for Mod, Ref, Document, and Set value types

ptpaterson:
## Problem
Once changes to the wire protocol for @ref, @doc, and @set are deployed, we can provide custom classes that will encode back to the tagged wire format, but which users don't have to think about the tags. Also this gives us a place to put additional functionality (like iterators/pagination helpers on the `Set` class) if we want/need it.

## Solution
Provide a base `DocumentRef` and `NamedDocumentRef` classes which `@ref` will decode into.

`DocumentRef` and `NamedDocumentRef` depends on a `Mod` class, which `@mod` will decode into.

Provide `Document` and `NamedDocument` classes which inherits from the `DocumentRef` and `NamedDocumentRef`, respectively. These are what `@doc` will decode into. They represent a materialized document. It can eventually be used as a base class for users implementing their own data types. They can have methods on it specific to any Document.

All top level fields from Fauna are added to `Document` and `NamedDocument`. Typescript users will need to cast them to appropriate types to have typed access to user-defined fields. I've provided a type alias `type DocumentT<T> = Document & T` for convenience.

Actually decoding and encoding of the new types.

## Out of scope
iterator methods for Set.

## Testing
Added some very basic tests to construct the Ref and Document type.


src/values/doc.ts
```diff
@@ -0,0 +1,85 @@
+import { JSONObject } from "../wire-protocol";
+import { TimeStub } from "./date-time";
+
+/**
+ * A Document Reference. Represents a pointer to a Document in the database. The
+ * Document may or may not exists.
+ */
+export class Ref {
```
ptpaterson, line 8: This doesn't differentiate between "named" references (e.g. Collections, Databases, Functions, Roles, AccessProviders, etc.) which are shaped like 

```json
{ "@ref": { "name": "Things", "coll": { "@mod": "Collection" } } }
```

ptpaterson: Can we apply a discriminated union type on top of `NamedDocument` to cover Collections vs Databases etc. or should we have subclasses for each?

ptpaterson: Also how do we handle user data on NamedDocument? Should that be generic?

ptpaterson: Started over from scratch. `Document` now always has the user-defined data at the top field, it just can't be used by typescript. Typescript users can cast to the provided `DocumentT` alias to extend `Document` with generic data.

src/values/doc.ts
```diff
@@ -0,0 +1,188 @@
...
+ * @example
+ * ```javascript
+ *  const thingRef = await client.query(fql`
+ *    Thing("101")
+ *  `);
+ *
+ *  const id = thingRef.id
+ * ```
+ */
+export class DocumentRef {
```
cleve-fauna, line 18: let's name this `DocumentReference` to conform to what we're doing in Python.

src/values/doc.ts
```diff
@@ -0,0 +1,188 @@
...
+ * @example
+ * ```javascript
+ *  const thingCollection = await client.query(fql`
+ *    Thing.definition
+ *  `);
+ *
+ *  const id = thingCollection.id
+ * ```
+ */
+export class NamedDocumentRef {
```
cleve-fauna, line 81: Let's name this `NamedDocumentReference` to conform with Python

src/values/doc.ts
```diff
@@ -0,0 +1,188 @@
...
+ * @example
+ * ```javascript
+ *  const thingModule = await client.query(fql`
+ *    Thing
+ *  `);
+ *
+ *  const name = thingModule.name
+ * ```
+ */
+export class Mod {
```
cleve-fauna, line 160: Let's name this `Module` to conform with Python

src/values/doc.ts
```diff
@@ -0,0 +1,193 @@
...
+    if (typeof coll === "string") {
+      this.coll = new Module(coll);
+    } else {
+      this.coll = coll;
+    }
+  }
+}
+
+/**
+ * A Materialized Document with a name. AccessProviders, Collections, Databases,
```
cleve-fauna, line 102:
```suggestion
 * A materialized Document with a name. AccessProviders, Collections, Databases,
```

src/values/doc.ts
```diff
@@ -0,0 +1,193 @@
...
+    Object.assign(this, rest);
+  }
+
+  toObject(): { coll: Module; id: string; ts: TimeStub } {
+    return { ...this } as { coll: Module; id: string; ts: TimeStub };
+  }
+}
+
+/**
+ * A Reference to a Document with a name. The Document may or may not exists.
```
cleve-fauna, line 73:
```suggestion
 * A reference to a Document with a name. The Document may or may not exists.
```

src/values/doc.ts
```diff
@@ -0,0 +1,193 @@
...
+    if (typeof coll === "string") {
+      this.coll = new Module(coll);
+    } else {
+      this.coll = coll;
+    }
+  }
+}
+
+/**
+ * A Materialized Document with an ID. Keys, Tokens and Documents in
```
cleve-fauna, line 34:
```suggestion
 * A materialized Document with an ID. Keys, Tokens and Documents in
```

src/values/doc.ts
```diff
@@ -0,0 +1,193 @@
+import { JSONObject } from "../wire-protocol";
+import { TimeStub } from "./date-time";
+
+/**
+ * A Reference to a Document with an ID. The Document may or may not exists.
```
cleve-fauna, line 5:
```suggestion
 * A reference to a Document with an ID. The Document may or may not exists.
```

src/values/doc.ts
```diff
@@ -0,0 +1,193 @@
+import { JSONObject } from "../wire-protocol";
+import { TimeStub } from "./date-time";
+
+/**
+ * A reference to a Document with an ID. The Document may or may not exists.
```
cleve-fauna, line 5:
```suggestion
 * A reference to a Document with an ID. The Document may or may not exist.
```

src/values/doc.ts
```diff
@@ -0,0 +1,193 @@
...
+    Object.assign(this, rest);
+  }
+
+  toObject(): { coll: Module; id: string; ts: TimeStub } {
+    return { ...this } as { coll: Module; id: string; ts: TimeStub };
+  }
+}
+
+/**
+ * A reference to a Document with a name. The Document may or may not exists.
```
cleve-fauna, line 73:
```suggestion
 * A reference to a Document with a name. The Document may or may not exist.
```

src/values/doc.ts
```diff
@@ -0,0 +1,193 @@
+import { JSONObject } from "../wire-protocol";
+import { TimeStub } from "./date-time";
+
+/**
+ * A reference to a Document with an ID. The Document may or may not exist.
+ * References to Keys, Tokens, and Documents in user-defined Collections are
+ * decoded into a {@link DocumentReference}.
```
cleve-fauna, line 7:
```suggestion
 * References to Keys, Tokens, and Documents in user-defined Collections are
 * modeled with a {@link DocumentReference}.
```

src/values/doc.ts
```diff
@@ -0,0 +1,193 @@
...
+  readonly name: string;
+
+  constructor(name: string) {
+    this.name = name;
+  }
+}
+
+/**
+ * A Document extended with user-defined data. Typescript users can cast
+ * instances of {@link Document} to {@link DocumentT} to access user-defined fields with type safety.
```
cleve-fauna, line 175:
```suggestion
 * A Document of a Collection typed with a user-defined data type. Typescript users can cast
 * instances of {@link Document} to {@link DocumentT} to access user-defined fields with type safety.
```

cleve-fauna (approved):
@ptpaterson it would be good to add round-trip tests for every data type we support.

Tests that instantiate every JS type we support getting in and out of Fauna - and then put it in Fauna and get it back out. To prove that we get it out in the same shape.

Fine with me to cover that in a different PR

PR #91, opened 2023-03-13
FE-3150 Add constraint_failures to wire protocol.

cleve-fauna:
Ticket(s): FE-3150

## Problem
- While fixing dashboard bugs noticed we are not including `constraint_failures` in our wire protocol.
- Same is true across all drivers
  - https://github.com/fauna/fauna-go/blob/fqlx-beta/error.go
  - https://github.com/fauna/fauna-python/blob/main/fauna/errors.py

## Solution
- Include optional constraint_failures array on `QueryFailure` wire model

## Result
- clients, such as the dashboard, can detect constraint_failures

## Testing
Verifying tests still pass.

### DIFF
--- a/__tests__/integration/query.test.ts
+++ b/__tests__/integration/query.test.ts
@@ -1,10 +1,6 @@
-import { env } from "process";
 import { getClient } from "../client";
 import { Client } from "../../src/client";
-import {
-  type ClientConfiguration,
-  endpoints,
-} from "../../src/client-configuration";
+import { type ClientConfiguration } from "../../src/client-configuration";
 import {
   AuthenticationError,
   ClientError,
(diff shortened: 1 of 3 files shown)


ptpaterson (approved):
Makes me wonder if we should provide a more specific type, like including all of the possible codes as a type, instead of just `string`. Then we could make this a union type where the QueryFailure is either a constraint failure with the extra data or the other errors without it.

I'd love to hear what you think about that, but I am totally good with this as is.

cleve-fauna:
> Makes me wonder if we should provide a more specific type, like including all of the possible codes as a type, instead of just `string`. Then we could make this a union type where the QueryFailure is either a constraint failure with the extra data or the other errors without it.
> 
> I'd love to hear what you think about that, but I am totally good with this as is.

Concern with that would be we couple ourselves to the database release cycle / we get brittle if they decide to change or add anything.

cleve-fauna:
If anything @ptpaterson i was thinking we might want to lean towards eliminating the distinction between a QueryCheckError and a QueryRuntimeError as it really doesn't matter from a user's perspective - the query is bad either way. So we could rebundle as QueryError or some such.

pnwpedro (approved):

PR #92, opened 2023-03-16
Add typecheck configuration option

ptpaterson:
Ticket(s): [FE-3181](https://faunadb.atlassian.net/browse/FE-3181)

## Problem
We need to expose an option to enable or disable typechecking at the Client and the Query. The default should be that x-typecheck is unset.

## Solution
- Add an optional `typecheck` field to `ClientConfiguration`
- Add an optional `typecheck` field to `QueryRequestHeaders`

## Testing
Extended tests to ensure 
- the client configuration sets a header; and 
- query options sets a header and overrides the client configuration

[FE-3181]: https://faunadb.atlassian.net/browse/FE-3181?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -64,7 +64,8 @@ export interface ClientConfiguration {
   traceparent?: string;
   /**
    * Enable or disable typechecking of the query before evaluation. If no value
-   * is provided, the database defaults using typechecking.
+   * is provided, the value of `typechecked` in the database configuration will
+   * be used.
    */
   typecheck?: boolean;
 }
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -62,6 +62,11 @@ export interface ClientConfiguration {
    * Must match format: https://www.w3.org/TR/trace-context/#traceparent-header
    */
   traceparent?: string;
+  /**
+   * Enable or disable typechecking of the query before evaluation. If no value
+   * is provided, the database defaults using typechecking.
+   */
+  typecheck?: boolean;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/client-configuration.ts
```diff
@@ -62,6 +62,11 @@ export interface ClientConfiguration {
    * Must match format: https://www.w3.org/TR/trace-context/#traceparent-header
    */
   traceparent?: string;
+  /**
+   * Enable or disable typechecking of the query before evaluation. If no value
+   * is provided, the database defaults using typechecking.
```
pnwpedro, line 67: I think it defaults to using the database's setting, so let's reword a bit. https://github.com/fauna/core/blob/master/ext/api/src/main/resources/openapi.yml#L73-L75

```suggestion
   * Enable or disable typechecking of the query before evaluation. If no value
   * is provided, the `typechecked` setting on the database will be used.
```

src/wire-protocol.ts
```diff
@@ -48,6 +48,11 @@ export interface QueryRequestHeaders {
    * Overrides the optional setting for the client.
    */
   traceparent?: string;
+  /**
+   * Enable or disable typechecking of the query before evaluation. If no value
+   * is provided, the database defaults using typechecking.
```
pnwpedro, line 53: Similar comment

pnwpedro (changes requested):

PR #94, opened 2023-03-20
Fix QueryStats and QueryInfo types and tests

ptpaterson:
Ticket(s): [FE-3109](https://faunadb.atlassian.net/browse/FE-3109)

## Problem
In `QueryStats`, `storage_bytes_written` should be `storage_bytes_write`

The fields for `QueryInfo` are incorrectly specified as required/optional.

`QueryInfo.query_tags` is not decoded correctly. It comes back exactly how the header is sent, e.g. 

```
"project=teapot,hello=world,testing=foobar"
```

## Solution
Align types with the core spec.

Add decoding of `query_tags` if received.

## Testing
Added new test for `query_tags`.

[FE-3109]: https://faunadb.atlassian.net/browse/FE-3109?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/wire-protocol.ts
+++ b/src/wire-protocol.ts
@@ -78,13 +78,13 @@ export type QueryStats = {
 
 export type QueryInfo = {
   /** The last transaction timestamp of the query. A Unix epoch in microseconds. */
-  txn_ts: number;
+  txn_ts?: number;
   /** A readable summary of any warnings or logs emitted by the query. */
-  summary: string;
+  summary?: string;
   /** The value of the x-query-tags header, if it was provided. */
   query_tags?: Record<string, string>;
   /** Stats on query performance and cost */
-  stats: QueryStats;
+  stats?: QueryStats;
 };
 
 export type QuerySuccess<T> = QueryInfo & {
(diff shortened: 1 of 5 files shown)


src/wire-protocol.ts
```diff
@@ -80,9 +80,9 @@ export type QueryInfo = {
   /** The last transaction timestamp of the query. A Unix epoch in microseconds. */
   txn_ts: number;
   /** A readable summary of any warnings or logs emitted by the query. */
-  summary?: string;
+  summary: string;
```
cleve-fauna, line 83: Looks to me like all fields in `QueryInfo` are optional:

https://github.com/fauna/core/blob/master/ext/api/src/main/resources/openapi.yml#L353

cleve-fauna, line 83: Looks to me like all fields in `QueryInfo` are optional:

https://github.com/fauna/core/blob/master/ext/api/src/main/resources/openapi.yml#L353

ptpaterson: Oh, right, it's only definitely there with query success. Will make all of them optional.

cleve-fauna (approved):

cleve-fauna (approved):

PR #96, opened 2023-03-20
Commit for dashboard development

cleve-fauna:
- Support decorated format. Fix linting errors
- Commit to put build in the repo to allow the dashboard to use latest constructs
- Commit to put build in the repo to allow the dashboard to use latest constructs

Leave this as is until we are in NPM.

Needed for dashboard development.


cleve-fauna:
Was missing needed models

PR #99, opened 2023-03-22
Test domain types and fix encoding

ptpaterson:
Ticket(s): [FE-3214](https://faunadb.atlassian.net/browse/FE-3214)

## Problem
Did not have tests for domain types and were finding bugs.

- Query input and output are restricted to the `JSONValue` type, but the new classes are not included in it.
- Document and NamedDocument are incorrectly encoded as DocumentReference and NamedDocumentReference.
- Core cannot (yet?) receive values tagged as `@set`

## Solution

- Add new domain classes to `JSONValue`.
- Fix encoding for documents.
- Encode Set by unwrapping it

## Testing
Added tests for all of the new classes.

### DIFF
--- a/src/tagged-type.ts
+++ b/src/tagged-type.ts
@@ -88,7 +88,8 @@ type TaggedLong = { "@long": string };
 type TaggedMod = { "@mod": string };
 type TaggedObject = { "@object": JSONObject };
 type TaggedRef = { "@ref": TaggedRefBase };
-type TaggedSet = { "@set": { data: JSONValue[]; after?: string } };
+// WIP: core does not accept `@set` tagged values
+// type TaggedSet = { "@set": { data: JSONValue[]; after?: string } };
 type TaggedTime = { "@time": string };
 
 export const LONG_MIN = BigInt("-9223372036854775808");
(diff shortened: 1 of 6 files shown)


src/tagged-type.ts
```diff
@@ -153,17 +157,20 @@ const encodeMap = {
   documentReference: (value: DocumentReference): TaggedRef => ({
     "@ref": { id: value.id, coll: { "@mod": value.coll.name } },
   }),
-  document: (value: Document): TaggedRef => ({
-    "@ref": { id: value.id, coll: { "@mod": value.coll.name } },
+  document: (value: Document): TaggedDoc => ({
+    "@doc": { id: value.id, coll: { "@mod": value.coll.name } },
```
pnwpedro, line 161: I'll follow with this in python, but originally this was intentional. It feels awkward to encode to @doc when it's really a ref (and lacks all of the data and the ts). @ref feels like it communicates more clearly what is going on.

ptpaterson: In that case, I will change back to encoding Document and NamedDocument to `@ref` and also expect that on round-trip back to the client that `@ref` results in a Document.

ptpaterson: ... unless the document doesn't exist. In that case, it is returned as `@ref` again.

I've reverted the encoding for docs back to using `@ref` and updated tests, including making it two tests, one for doc exists and one for when it doesn't.

PR #108, opened 2023-03-24
Client accepts only fql output as input

ptpaterson:
Ticket(s): [FE-3301](https://faunadb.atlassian.net/browse/FE-3301)

## Problem
We want to push users toward best practice by requiring them to use the `fql` function.

## Solution
Eliminate the `QueryBuilder` interface. Rename `TemplateQueryBuilder` to `Query`. Check if arguments are also `Query` instances with `instanceof` instead of a predicate function to check for the interface.

Client.query only accepts a `Query` instance.

## Testing
Removed all tests that specifically intended to exercise passing `QueryRequest` to `Client.query`. Fixed other tests to use `fql` function.

----

[FE-3301]: https://faunadb.atlassian.net/browse/FE-3301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -113,7 +113,7 @@ export class Client {
 
   /**
    * Queries Fauna.
-   * @param request - a {@link QueryRequest} or {@link Query} to build a request with.
+   * @param request - a {@link Query} to execute in Fauna.
    *  Note, you can embed header fields in this object; if you do that there's no need to
    *  pass the headers parameter.
    * @param headers - optional {@link QueryRequestHeaders} to apply on top of the request input.
--- a/src/client.ts
+++ b/src/client.ts
@@ -117,7 +117,7 @@ export class Client {
    *  Note, you can embed header fields in this object; if you do that there's no need to
    *  pass the headers parameter.
    * @param headers - optional {@link QueryRequestHeaders} to apply on top of the request input.
-   *   Values in this headers parameter take precedence over the same values in the request
+   *   Values in this headers parameter take precedence over the same values in the {@link ClientConfiguration}.
    *   parameter. This field is primarily intended to be used when you pass a Query as
    *   the parameter.
    * @returns Promise&lt;{@link QuerySuccess}&gt;.
--- a/src/client.ts
+++ b/src/client.ts
@@ -13,7 +13,7 @@ import {
   ServiceTimeoutError,
   ThrottlingError,
 } from "./errors";
-import type { QueryBuilder } from "./query-builder";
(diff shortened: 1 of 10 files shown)


src/client.ts
```diff
@@ -113,12 +113,12 @@ export class Client {
 
   /**
    * Queries Fauna.
-   * @param request - a {@link QueryRequest} or {@link QueryBuilder} to build a request with.
+   * @param request - a {@link QueryRequest} or {@link Query} to build a request with.
```
cleve-fauna, line 116:
```suggestion
   * @param request - a {@link Query} to execute in Fauna.
```

ptpaterson: Oh good catch on doc comments! 🙏🏻

src/client.ts
```diff
@@ -113,12 +113,12 @@ export class Client {
 
   /**
    * Queries Fauna.
-   * @param request - a {@link QueryRequest} or {@link QueryBuilder} to build a request with.
+   * @param request - a {@link Query} to execute in Fauna.
    *  Note, you can embed header fields in this object; if you do that there's no need to
    *  pass the headers parameter.
    * @param headers - optional {@link QueryRequestHeaders} to apply on top of the request input.
    *   Values in this headers parameter take precedence over the same values in the request
```
cleve-fauna, line 120:
```suggestion
   *   Values in this headers parameter take precedence over the same values in the {@link ClientConfiguration}
```

src/client.ts
```diff
@@ -113,12 +113,12 @@ export class Client {
...
-   * @param request - a {@link QueryRequest} or {@link QueryBuilder} to build a request with.
+   * @param request - a {@link Query} to execute in Fauna.
    *  Note, you can embed header fields in this object; if you do that there's no need to
    *  pass the headers parameter.
    * @param headers - optional {@link QueryRequestHeaders} to apply on top of the request input.
-   *   Values in this headers parameter take precedence over the same values in the request
-   *   parameter. This field is primarily intended to be used when you pass a QueryBuilder as
+   *   Values in this headers parameter take precedence over the same values in the {@link ClientConfiguration}.
+   *   parameter. This field is primarily intended to be used when you pass a Query as
    *   the parameter.
```
cleve-fauna, line 122:
```suggestion
```

cleve-fauna (approved):
We'll need to update the demo package as well. i've added that to our hit list.

PR #111, opened 2023-03-24
Client can accept endpoints with or without a trailing Slash

ptpaterson:
Ticket(s): [FE-3293](https://faunadb.atlassian.net/browse/FE-3293)

## Problem
Providing an endpoint without the trailing slash doesn't work

## Solution
Use the `URL` API to append `/query/1` to the endpoint.

## Result
final URL is the same with or without a trailing slash in the endpoint.

## Testing
Added new client configuration test.

----

[FE-3293]: https://faunadb.atlassian.net/browse/FE-3293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -75,7 +75,10 @@ export class Client {
       ...clientConfiguration,
       secret: this.#getSecret(clientConfiguration),
     };
-    this.#url = `${this.clientConfiguration.endpoint.toString()}query/1`;
+    this.#url = new URL(
+      "/query/1",
+      this.clientConfiguration.endpoint
+    ).toString();
     if (!httpClient) {
       this.#httpClient = getDefaultHTTPClient();
     } else {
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -75,7 +75,10 @@ export class Client {
       ...clientConfiguration,
       secret: this.#getSecret(clientConfiguration),
     };
-    this.#url = `${this.clientConfiguration.endpoint.toString()}query/1`;
+    this.#url = new URL(
```
cleve-fauna, line 78: will this work in node without an import?

ptpaterson: Yes! `URL` is built in to browsers and node etc.

ptpaterson: https://developer.mozilla.org/en-US/docs/Web/API/URL/URL#browser_compatibility

cleve-fauna: looks like its global: https://nodejs.org/api/globals.html#url so yes.

ptpaterson: Oh but valid question if it's available in the global scope. The answer to that is also yes.

cleve-fauna (approved):

src/client.ts
ptpaterson: haha beat me to it 👍🏻

PR #124, opened 2023-03-30
Close with reference counting to protect the global node http2 client

cleve-fauna:
## Problem
- customers need a way to close their clients

## Solution
- expose a close method
- for the shared HTTP/2 client in node use reference counting to only close when no more processes are using the global http2 connection.
## Result
- close available to users

## Testing

- full suite
- added tests to show we're tracking references, can recreate on global pool after all are closed.

----

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -2,6 +2,7 @@ import { ClientConfiguration, endpoints } from "./client-configuration";
 import {
   AuthenticationError,
   AuthorizationError,
+  ClientClosedError,
   ClientError,
   NetworkError,
   ProtocolError,
@@ -49,6 +50,8 @@ export class Client {
   #lastTxnTs?: number;
   /** url of Fauna */
   #url: string;
+  /** true if this client is closed false otherwise */
+  #isClosed = false;
 
   /**
    * Constructs a new {@link Client}.
@@ -115,6 +118,20 @@ export class Client {
   }
 
   /**
+   * Closes the underlying HTTP client. Subsquent query calls
+   * will fail.
+   */
+  close() {
+    if (this.#isClosed) {
+      throw new ClientClosedError(
+        "Your client is closed. You cannot close it again."
+      );
+    }
+    this.#httpClient.close();
+    this.#isClosed = true;
+  }
+
+  /**
    * Queries Fauna.
    * @param request - a {@link Query} to execute in Fauna.
    *  Note, you can embed header fields in this object; if you do that there's no need to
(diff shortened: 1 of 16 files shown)


src/http-client/node-http2-client.ts
```diff
@@ -50,6 +55,16 @@ export class NodeHTTP2Client implements HTTPClient {
     }
   }
 
+  /** {@inheritDoc HTTPClient.close} */
+  close() {
+    this.#numberOfUsers--;
+    if (this.#numberOfUsers === 0) {
```
cleve-fauna, line 61: could probably harden this to make sure we cannot go negative.

sprsquish (commented):
Is there a chance someone might close all instances and then need to open another one? I don't know this driver well enough that I can tell if something would blow up if I closed everything then tried to open another one.

cleve-fauna:
> Is there a chance someone might close all instances and then need to open another one? I don't know this driver well enough that I can tell if something would blow up if I closed everything then tried to open another one.

Good question @sprsquish . It should handle that case fine.

https://github.com/fauna/fauna-js/blob/bce968e915f47a2f4d98ca1407fce52f70e4e3d4/src/http-client/node-http2-client.ts#L75

But i will add a test for that.

src/http-client/index.ts
```diff
@@ -40,6 +40,18 @@ export interface HTTPClient {
    * @returns A Promise&lt;{@link HTTPResponse}&gt;
    */
   request(req: HTTPRequest): Promise<HTTPResponse>;
+
+  /**
+   * Flags the calling {@link Client} as no longer
+   * referencing this HTTPClient. Once no processes reference this client
```
cleve-fauna, line 46:
```suggestion
   * referencing this HTTPClient. Once no {@link Client} instances reference this HTTPClient
```

src/http-client/node-http2-client.ts
```diff
@@ -50,6 +53,28 @@ export class NodeHTTP2Client implements HTTPClient {
     }
   }
 
+  /** {@inheritDoc HTTPClient.close} */
+  close() {
+    if (this.isClosed()) {
+      return;
+    }
+    this.#numberOfUsers--;
+    // defend against redundant close calls
```
cleve-fauna, line 62:
```suggestion
    // defend against redundant close calls
    if (this.isClosed()) {
      return;
    }
    this.#numberOfUsers--;
```

src/http-client/node-http2-client.ts
```diff
@@ -50,6 +53,28 @@ export class NodeHTTP2Client implements HTTPClient {
...
 
+  /** {@inheritDoc HTTPClient.close} */
+  close() {
+    // defend against redundant close calls
+    if (this.isClosed()) {
+      return;
+    }
+    this.#numberOfUsers--;
+    if (this.#numberOfUsers === 0) {
+      this.#numberOfUsers = 0;
```
cleve-fauna, line 64:
```suggestion
```

src/client.ts
```diff
@@ -114,6 +117,20 @@ export class Client {
     return rest;
   }
 
+  /**
+   * Closes the underlying HTTP client. Subsquent query calls
```
cleve-fauna, line 121:
```suggestion
   * Closes the underlying HTTP client. Subsquent query or close calls
```

sprsquish (approved):

PR #126, opened 2023-04-07
Send x-format header by default

ptpaterson:
Ticket(s): [FE-3396](https://faunadb.atlassian.net/browse/FE-3396)

## Problem
In order to make sure that we are providing the best driverless-experience, we need to make sure that drivers are setting the format header so that the default can be the most appropriate for direct API access

## Solution
Make `format` a required client configuration field. Provide a default format.

## Result
Since the client configuration will always have a format, the `x-format` header will always be sent

## Out of scope
N/A

## Testing
tweaked the headers test so that it uses the default client configuration for format (i.e. not explicitly set), so we now test the the header is sent by default.

----

[FE-3396]: https://faunadb.atlassian.net/browse/FE-3396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -11,19 +11,18 @@ export interface ClientConfiguration {
   /**
    * Determines the encoded format expected for the query `arguments` field, and
    * the `data` field of a successful response.
-   * @remarks **Note, it is very unlikely you need to change this value from its default.**
-   * By default the driver transmits type information over the wire. Fauna also assumes type information is
-   * transmitted by default and thus leaving this value undefined will allow Fauna and the driver to send and
-   * receive type data.
-   *  Type information allows the driver and Fauna to distinguish between types such as int" and "long" which do not
+   * @remarks **Note, it is very unlikely you need to change this value from its
+   * default.**
+   * The default format is "tagged", which specifies that the driver transmits
+   * type information over the wire. Type information allows the driver and
+   * Fauna to distinguish between types such as int" and "long" which do not
    * have a standard way of distinguishing in JSON.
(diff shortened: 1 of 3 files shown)


cynicaljoy:
@ptpaterson my understanding of the requirement was that we hardcode it. I don't think it needs to be in the `ClientConfiguration`

ptpaterson:
hard coded default but that the user can configure if desired. We want to use the decorated format in the dashboard for example. I don't think Matt is saying the drivers are disallowed from setting the format.

cynicaljoy:
I just want to get a consensus on the expected user experience. We don't provide obvious configuration options in the other drivers, though there are escape hatches that a user could override the setting with.

src/client-configuration.ts
```diff
@@ -25,7 +25,7 @@ export interface ClientConfiguration {
    * It is not recommended that users use the "simple" format as you will lose the typing of your data. e.g. a "Date"
    * will no longer be recognized by the Fauna as a "Date", but will instead be treated as a string.
    */
-  format?: ValueFormat;
+  format: ValueFormat;
```
cleve-fauna, line 28: please update the remakrs above as we no longer needs lines 15-17

cleve-fauna (approved):
Please update the docs but looks fine.

src/client-configuration.ts
ptpaterson: Updated remarks.

cleve-fauna (approved):

PR #127, opened 2023-04-11
Implement string format for @set

ptpaterson:
Ticket(s): [FE-3394](https://faunadb.atlassian.net/browse/FE-3394)

## Problem
There is an edge case with set materialization which could result in materializing the tree to an arbitrary depth, or inifinitely recurse (specifically in this case if a set contains itself as an element). 

In order to solve this we need to introduce a variant to the @set tag where the value can be a raw cursor string. 

Before: @set value is an object: { data: Array<T>, after: string | null } 

After: @set value is a union: string | { data: Array<T>, after: string | null }

## Solution
Make data optional and update encoding & decoding.

## Out of scope
encoding Page into a `@set` tagged value. Core doesn't accept `@set` tagged values yet.

## Testing
Added encoding & decoding tests for Page. 

----

[FE-3394]: https://faunadb.atlassian.net/browse/FE-3394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/tagged-type.ts
```diff
@@ -165,12 +168,20 @@ const encodeMap = {
...
-    data: encodeMap["array"](value.data),
-    after: value.after,
-  }),
+  set: (value: Page<any>) => {
+    if (value.data === undefined) {
+      // if a Page has no data, then it must still have an 'after' cursor
+      // WIP: core does not accept `@set` tagged values, yet, so just unwrap for now
+      // return { "@set": value.after }
+      return value.after as string;
+    }
```
pnwpedro, line 177: I'm not sure if we should be encoding `Page` at all until core accepts `@set`. It might create more confusion than anything, especially this string case.

If we must encode it, and it will eventually be `{ @set: value.after }`, then this is prob the right choice

ptpaterson: Ah you are right, the semantics of sending a set to Fauna will change once they start handling it. I agree it will be better to just not handle encoding Pages yet. We can throw an error and tell folks to use the data and after fields directly instead.

pnwpedro (approved):

PR #128, opened 2023-04-12
Add AbortError, InvalidRequestError, and ContendedTransactionError classes

ptpaterson:
Ticket(s): 
- [FE-3388](https://faunadb.atlassian.net/browse/FE-3388)
- [FE-3402](https://faunadb.atlassian.net/browse/FE-3402)

## Problem
We should distinguish between 400 errors from invalid reuqests, calling `abort`, and those from other issues, so users can more easily handle them.

We've not added an error for 409 errors yet.

## Solution
Create a AbortError class and throw it if a 400 error has the code "abort".
Create an InvalidRequestError class and throw it if a 400 error has the code "invalid_request".
Create a ContendedTransactionError class and throw it on a 409 response.

Also included a nit about using `expect.toBe` instead of `expect.toEqual` which is more for asserting the values of objects.

## Out of scope
n/a

## Testing
Core updated the wire protocol, so we can now also test that the code coming back is "abort" and not "aborted".

----

[FE-3388]: https://faunadb.atlassian.net/browse/FE-3388?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FE-3402]: https://faunadb.atlassian.net/browse/FE-3402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/integration/query.test.ts
+++ b/__tests__/integration/query.test.ts
@@ -47,7 +47,7 @@ describe("query", () => {
   it("Can query an FQL-x endpoint", async () => {
     const result = await client.query<number>(fql`"taco".length`);
 
-    expect(result.data).toEqual(4);
+    expect(result.data).toBe(4);
     expect(result.summary).toBeDefined();
     expect(result.txn_ts).toBeDefined();
     expect(result.stats).toBeDefined();
@@ -63,7 +63,7 @@ describe("query", () => {
   it("Can query with arguments", async () => {
     const str = "taco";
     const result = await client.query(fql`${str}.length`);
-    expect(result.data).toEqual(4);
+    expect(result.data).toBe(4);
     expect(result.txn_ts).toBeDefined();
   });
 
@@ -116,9 +116,7 @@ describe("query", () => {
       const httpClient: HTTPClient = {
         async request(req) {
           Object.entries(expectedHeaders).forEach(([_, expectedHeader]) => {
-            expect(req.headers[expectedHeader.key]).toEqual(
-              expectedHeader.value
-            );
+            expect(req.headers[expectedHeader.key]).toBe(expectedHeader.value);
           });
           return dummyResponse;
         },
@@ -145,7 +143,7 @@ describe("query", () => {
       await client.query(fql`happy little fox`);
     } catch (e) {
       if (e instanceof QueryCheckError) {
-        expect(e.httpStatus).toEqual(400);
+        expect(e.httpStatus).toBe(400);
         expect(e.message).toBeDefined();
         expect(e.code).toBeDefined();
         expect(e.queryInfo?.summary).toBeDefined();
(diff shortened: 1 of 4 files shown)


src/errors.ts
```diff
@@ -92,15 +85,37 @@ export class QueryRuntimeError extends ServiceError {
...
     }
     this.name = "QueryCheckError";
   }
 }
 
+/**
+ * An error due to a "compile-time" check of the query
+ * failing.
+ */
+export class QueryAbortError extends ServiceError {
```
pnwpedro, line 101: Maybe just AbortError so it's consistent with openapi? I don't think we should diverge from their naming conventions where possible.

```suggestion
export class AbortError extends ServiceError {
```

src/errors.ts
```diff
@@ -92,15 +85,37 @@ export class QueryRuntimeError extends ServiceError {
...
+    super(failure, 400);
     if (Error.captureStackTrace) {
       Error.captureStackTrace(this, QueryCheckError);
     }
     this.name = "QueryCheckError";
   }
 }
 
+/**
+ * An error due to a "compile-time" check of the query
```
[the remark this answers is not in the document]
ptpaterson: also the AbortError is not a compile time error. That was a copy-paste error.

ptpaterson:
Will add `InvalidRequest` as well.

__tests__/integration/query.test.ts
```diff
@@ -62,7 +69,7 @@ describe("query", () => {
   it("Can query with arguments", async () => {
     const str = "taco";
     const result = await client.query(fql`${str}.length`);
-    expect(result.data).toEqual(4);
```
cleve-fauna, line 65: why did we change all these? just curious

cleve-fauna (dismissed):
Please update the exports https://github.com/fauna/fauna-js/blob/main/src/index.ts#L7

or these will not be useable

__tests__/integration/query.test.ts
ptpaterson: A nit pick. `.toBe` is supposed to be for comparing literal values and `.toEqual` is for comparing objects.

cleve-fauna: my understanding is toBe compares refernces (addresses in memory) and toEqual compares values.

I don't think there's really any reason for us to care about the references in memory of these types of things.

src/client.ts
```diff
@@ -222,26 +226,35 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+        return new ThrottlingError(failure);
       case 440:
-        return new QueryTimeoutError(failure, httpStatus);
+        return new QueryTimeoutError(failure);
       case 500:
-        return new ServiceInternalError(failure, httpStatus);
+        return new ServiceInternalError(failure);
       case 503:
-        return new ServiceTimeoutError(failure, httpStatus);
+        return new ServiceTimeoutError(failure);
```
cleve-fauna, line 257: nit, but IMO this makes the code more brittle as now we're losing the type checking that we're passing the right error codes to the right errors.

__tests__/integration/query.test.ts
ptpaterson: `toBe` might compare the reference for Objects, but the docs say it is for asserting a value is what you expect.
image

src/client.ts
ptpaterson: That makes sense. I changed it back.

PR #134, opened 2023-04-17
Fix some errors in the doc.ts documentation and make the examples a bit easier to read.

cleve-fauna:
## Problem
There were a couple errors in the documentation in doc.ts that was using an older FQL X syntax.

## Solution
Revamp the documentation in this file to fix bugs and make the examples a bit more spelled out.

I also took the liberty of changing "Thing" to "Users" as it feels a bit more concrete.

## Result
Documentation is up to date.

## Testing
N/A
----

### DIFF
--- a/src/values/doc.ts
+++ b/src/values/doc.ts
@@ -207,7 +207,7 @@ export class Module {
  *    Users.byId("101")
  *  `);
  *
- *  const color = user.color
+ *  const color = user.color;
  * ```
  *
  * @remarks The {@link Document} class cannot be generic because classes cannot
--- a/src/values/doc.ts
+++ b/src/values/doc.ts
@@ -6,13 +6,17 @@ import { TimeStub } from "./date-time";
  * References to Keys, Tokens, and Documents in user-defined Collections are
  * modeled with a {@link DocumentReference}.
  *
+ * The example below retrieves a document reference from a
+ * hypothetical "Users" collection.
+ *
  * @example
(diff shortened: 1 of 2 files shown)


src/values/doc.ts
```diff
@@ -119,11 +133,11 @@ export class NamedDocumentReference {
...
  *  }
  *
- *  const thingCollection = await client.query<NamedDocument<CollectionMetadata>>(fql`
- *    Thing.definition
+ *  const userCollection = await client.query<NamedDocument<CollectionMetadata>>(fql`
+ *    Users.definition
  *  `);
  *
- *  const metadata = thingCollection.data.metadata
+ *  const metadata = userCollection.data.metadata
```
ptpaterson, line 140:
```suggestion
 *  const metadata = userCollection.data.metadata;
```

src/values/doc.ts
```diff
@@ -174,17 +194,20 @@ export class Module {
...
  *  }
  *
- *  const thing = await client.query<DocumentT<Thing>>(fql`
- *    Thing.byId("101")
+ *  const user = await client.query<DocumentT<User>>(fql`
+ *    Users.byId("101")
  *  `);
  *
- *  const color = thing.color
+ *  const color = user.color
```
ptpaterson, line 210:
```suggestion
 *  const color = user.color;
```

ptpaterson (approved):

PR #138, opened 2023-04-26
Make type parameter to Client.query more specific

ptpaterson:
## Problem
`Client.query` cannot actually return just `any`. The `QueryResponse` type will always contain a `QueryValue` after decoding is done.

## Solution
Add type bounds to `Client.query`

## Result
Typescript users cannot accidentally try to receive their own classes from `Client.query`.

## Out of scope
n/a

## Testing
n/a

----

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -159,7 +159,7 @@ export class Client {
    * @throws {@link ClientClosedError} if a query is issued after the client is closed.
    * due to an internal error.
    */
-  async query<T = any>(
+  async query<T extends QueryValue = any>(
     request: Query,
     headers?: QueryRequestHeaders
   ): Promise<QuerySuccess<T>> {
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -159,7 +159,7 @@ export class Client {
    * @throws {@link ClientClosedError} if a query is issued after the client is closed.
    * due to an internal error.
    */
-  async query<T = any>(
+  async query<T extends QueryValue = any>(
```
cleve-fauna, line 162: hypothetically a user could build their own types that match the output shape of the query API.

We can discuss the pros / cons of leaving this as it was to support that offline.

ptpaterson:
```suggestion
  async query<T extends QueryValue>(
```

After working more with the new `client.paginate` I don't think default type parameter is a good idea.

But yes, let's catch up offline.

cleve-fauna (approved):

PR #139, opened 2023-04-26
Pagination iterator type

ptpaterson:

## Problem

## Solution
Create a `PaginationHelper` class that implements `AsyncGenerator`.

Add a method to `Client` to create a `PaginationHelper` from a `Page`.

## Result
`PaginationHelper`s are created by calling `client_instance.paginate(page_instance)`.

Pages can be fetched using `for await (data of iterator) { ... }` syntax to asynchronously fetch all pages.

`iterable.next()` method can also be used to more control.

Added a `iterable.previous()` method as well to go back a page. Each call to `next` caches the value in an array, so `previous` will return the cached values.

## Out of scope
n/a

## Testing
Added new tests.

----

### DIFF
--- a/__tests__/integration/page.test.ts
+++ b/__tests__/integration/page.test.ts
@@ -1,4 +1,4 @@
-import { fql, Page } from "../../src";
+import { DocumentT, fql, Page } from "../../src";
 import { getClient } from "../client";
 
 const client = getClient({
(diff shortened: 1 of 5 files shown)


cleve-fauna:
would be more intuitive less verbose IMO if data itself was an iterable when we are hitting a pageable entity.

then customers could do something like:

```
if (respone.data.isPageable()) {
  for (let page of await response.data) {
    for (let item of page) {

    }
}
```

IIRC Luca's prototype did something like this.
}

PR #140, opened 2023-04-26
Add client arg to decode function

ptpaterson:

IMPORTANT: this is ONLY necessary if we go with #141 option for pagination iterator

## Problem
Decoding context has no client in it. If we want to decode results into classes that contain a client, then the decoder needs to have a client.

## Solution
add a client argument to the decode function 

## Result
`Client.query` is the only place that calls `TaggedTypeFormat.decode`; the client now passes in `this` to `decode`. The decode function is now ready to use the client where we need it.

## Out of scope
Implementing pagination helpers. This is a small step to building a Page class that can make additional calls to the database.

## Testing
Updated decoding tests where necessary.

----

### DIFF
--- a/__tests__/unit/tagged-format.test.ts
+++ b/__tests__/unit/tagged-format.test.ts
@@ -12,6 +12,9 @@ import {
   TaggedTypeFormat,
   TimeStub,
 } from "../../src";
+import { getClient } from "../client";
+
+const client = getClient();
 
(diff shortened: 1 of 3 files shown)


cleve-fauna:
If we go with #143 this can be closed?

ptpaterson:
Yes, this is ONLY necessary if we go with #141

PR #141, opened 2023-04-27
Pagination iterator type (1) - @set decodes to iterable with built-in client

ptpaterson:
depends on #140 

## Problem

We want to limit friction for users getting paginated data.

Users need to have control over remote fetches, and need access to the after-cursors if they want them.

## Solution
Create a `SetIterator` class that implements `AsyncGenerator`. `SetIterator` most closely represents a Fauna Set, which can be a materialized Set or an embedded Set. `Page` stays very simple. The `SetIterator` iterator returns instances of `Page`.

When the client receives an `@set` tagged value, it is decoded into the `SetIterator` class. It can be used like this

```typescript
const response = await client.query(fql`MyCollection.all()`);
const setIterator = response.data

for await (const page of setIterator) {
  // do stuff with page
}
```

The helper is constructed with a reference to the client so it can make additional calls to the database without any intermediate steps.

## Result
Pages can be fetched using `for await (page of pagination_helper) { ... }` syntax to asynchronously fetch all pages.

The AsyncGenerator methods `next()`, `return()`, and `throw()` can also be used to provide more control.

## Out of scope
n/a

## Testing
Added new tests.

----

### DIFF
--- a/__tests__/integration/page.test.ts
+++ b/__tests__/integration/page.test.ts
@@ -119,4 +119,36 @@ describe("PaginationHelper", () => {
       }
     }
   });
+
+  it("can can stop the iterator with the return method", async () => {
+    expect.assertions(1);
+
+    const response = await client.query<PaginationHelper<MyDoc>>(
+      fql`IterTestBig.all()`
+    );
+    const paginationHelper = response.data;
+
+    for await (const page of paginationHelper) {
+      expect(page.data.length).toBe(16);
+      await paginationHelper.return();
+    }
+  });
+
+  it("can can stop the iterator with the throw method", async () => {
+    expect.assertions(2);
+
+    const response = await client.query<PaginationHelper<MyDoc>>(
+      fql`IterTestBig.all()`
+    );
+    const paginationHelper = response.data;
+
+    try {
+      for await (const page of paginationHelper) {
+        expect(page.data.length).toBe(16);
+        await paginationHelper.throw("oops");
+      }
+    } catch (e) {
+      expect(e).toBe("oops");
+    }
+  });
 });
--- a/__tests__/integration/page.test.ts
+++ b/__tests__/integration/page.test.ts
@@ -1,5 +1,4 @@
-import { DocumentT, fql } from "../../src";
-import { PaginationHelper } from "../../src/values";
+import { DocumentT, fql, SetIterator } from "../../src";
 import { getClient } from "../client";
 
 const client = getClient({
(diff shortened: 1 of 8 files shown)


mwilde345:
Correct me if I'm wrong but using just the generator would require users to store pages in memory as they iterate, if they want to access a previous page. What are thoughts on wrapping up a prev() or something which keeps track of the current after token? We have something like that built into the dashboard right now, having moved away from a generator implementation. Thinking customers might do the same

ptpaterson:
The prior iteration (#139) included a built-in cache for previous pages. Immediate feedback was that caching all of that unbounded data is problematic. I proposed just storing the cursor, and/or only storing the previous N pages of data along with the cursors. We concluded that it was a bit much for the driver to handle and users can add on something themselves. They/We can do that as long as the iterator returns the whole page to provide access to the cursors at each step.

You bringing this up makes me feel less crazy for proposing it myself, though! 🙂 

The class in this PR could be extended with a cache like the following. Making the cache smarter, or only saving the cursors to refetch previous pages wouldn't be too much more difficult.

```typescript
export class PaginationHelperWithCache<
  T extends QueryValue
> extends PaginationHelper<T> {
  readonly #cache: Page<T>[] = [];
  #iter = -1;

  constructor(client: Client, initial: PageObject<T> | EmbeddedSet) {
    super(client, initial);
  }

  async next(): Promise<IteratorResult<Page<T>, void>> {
    console.log(`next() -- ${this.#iter} :: ${this.#cache.length}`)
    let next: IteratorResult<Page<T>, void>;
    if (this.#iter === this.#cache.length - 1) {
      next = await super.next();
      if (!next.done) {
        this.#cache.push(next.value);
        this.#iter++;
      }
    } else {
      this.#iter++;
      next = {
        done: false,
        value: this.#cache[this.#iter],
      };
    }

    return next;
  }

  async previous(): Promise<IteratorResult<Page<T>, void>> {
    if (this.#iter < 1) {
      return { done: true, value: undefined };
    }

    this.#iter--;
    return {
      done: false,
      value: this.#cache[this.#iter],
    };
  }
}
```

pnwpedro:
Nice! I love all of the thought you've put into pagination, and this impl feels pretty dang good.

Take this with a grain of salt because I'm not a JS dev, but I wonder if a slightly better API would be `new PaginationHelper<MyType>(client, query)` rather than building this as decoded type—or `PageIterator` or `FaunaSetIterator` or `<AnythingButHelper>` as a name because I'm not particularly in love with the ambiguity of Helper. 😁 

This allows the iterator to handle Pages, which are part of the wire protocol, without moonlighting as a wire protocol member itself. Second, the first query doesn't happen until you iterate, which makes it properly lazy. Finally, it should preclude any need for embedding the client in the encoder.

The trade off is the caller has to instantiate this, but I like this because I can setup my objects and then start querying when I iterate. This also allows me to handle errors in one place if I need to, rather than at the initial query and again inside my iterate code.

Anyway, I'd support what you have if consensus is there, just an alternative to consider!

PR #143, opened 2023-05-02
Pagination iterator type (3) - Can iterate directly over a query

ptpaterson:

## Problem

We want to limit friction for users getting paginated data.

Users need to have control over remote fetches, and need access to the after-cursors if they want them.

## Solution
Add the `EmbeddedSet` class to differentiate between a materialized Set (Page) and an unmaterialized one. This lets us enforce that a `Page` instance always contains data.

Add a `SetIterator` class. This version of `SetIterator` is different in a couple of ways:

- Instances of `Page` are unwrapped, so results from iterating are always `T[]` instead of `Page<T>`.
- `SetIterator` can be constructed from a nullary function that returns `Promise<QuerySuccess<QueryValue>>`. The first iteration will execute the function asynchronously. If the result is a `Page` then it will return the unwrapped data and continue iterating. If the result is not a `Page` then that value will be returned and iteration will stop.

Can be constructed with a query through the client

```typescript
const setIterator = client.paginate(fql`MyCollection.all()`);

for await (const page of setIterator) { ... }
```

constructed with an existing page through the client,

```typescript
const response = await client.query(fql`MyCollection.all()`);
const setIterator = client.paginate(page)

for await (const page of setIterator) { ... }
```

or still constructed directly,

```typescript
const response = await client.query(fql`MyCollection.all()`);
const page = response.data;
const setIterator = new SetIterator(client, page);

for await (const page of setIterator) { ... }
```

## Out of scope
n/a

## Testing
Added new tests.

----

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -137,6 +137,45 @@ export class Client {
     this.#isClosed = true;
   }
 
+  /**
+   * Creates an iterator to yield pages of data. If additional pages exist, the
+   * iterator will lazily fetch addition pages on each iteration.
+   *
+   * @typeParam T - The expected type of the items returned from Fauna on each
+   * iteration
+   * @param iterable - a {@link Query} or an existing fauna Set ({@link Page} or
+   * {@link EmbeddedSet})
+   * @returns A {@link SetIterator} that lazily fetches new pages of data on
+   * each iteration
+   *
+   * @example
+   * ```javascript
+   *  const userIterator = await client.paginate(fql`
+   *    Users.all()
+   *  `);
+   *
+   *  for await (const users of userIterator) {
+   *    for (const user of users) {
+   *      // do something with each user
+   *    }
+   *  }
+   * ```
+   *
+   * @example
+   * The {@link SetIterator.flatten} method can be used so the iterator yields
+   * items directly. Each item is fetched asynchronously and hides when
+   * additional pages are fetched.
+   *
+   * ```javascript
+   *  const userIterator = await client.paginate(fql`
+   *    Users.all()
+   *  `);
+   *
+   *  for await (const user of userIterator.flatten()) {
+   *    // do something with each user
+   *  }
+   * ```
+   */
   paginate<T extends QueryValue>(
     iterable: Page<T> | EmbeddedSet | Query
   ): SetIterator<T> {
@@ -148,12 +187,15 @@ export class Client {
 
   /**
    * Queries Fauna.
+   *
+   * @typeParam T - The expected type of the response from Fauna
    * @param request - a {@link Query} to execute in Fauna.
    *  Note, you can embed header fields in this object; if you do that there's no need to
    *  pass the headers parameter.
    * @param headers - optional {@link QueryRequestHeaders} to apply on top of the request input.
    *   Values in this headers parameter take precedence over the same values in the {@link ClientConfiguration}.
    * @returns Promise&lt;{@link QuerySuccess}&gt;.
+   *
    * @throws {@link ServiceError} Fauna emitted an error. The ServiceError will be
    *   one of ServiceError's child classes if the error can be further categorized,
    *   or a concrete ServiceError if it cannot. ServiceError child types are
--- a/src/client.ts
+++ b/src/client.ts
@@ -17,7 +17,7 @@ import {
   ContendedTransactionError,
   InvalidRequestError,
 } from "./errors";
-import type { Query } from "./query-builder";
+import { Query } from "./query-builder";
 import {
   isQueryFailure,
   isQuerySuccess,
@@ -33,7 +33,7 @@ import {
   type HTTPClient,
 } from "./http-client";
 import { TaggedTypeFormat } from "./tagged-type";
-import { Page, SetIterator } from "./values";
+import { EmbeddedSet, Page, SetIterator } from "./values";
 
 const defaultClientConfiguration: Pick<
   ClientConfiguration,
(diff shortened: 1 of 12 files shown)


ptpaterson:
Made some additional changes:

- Iterating a query was incorrectly returning the result of the query instead of an array containing the result. Fixed.
- Split out the logic for iterating actual sets and spoofing one if the query result is not a Page.
- Implemented an additional iterator (`FlattenedSetIterator`) that returns all of the items from all pages.
- Added a `flatten()` method to `SetIterator` to convert to `FlattenedSetIterator`)

src/client.ts
```diff
@@ -136,6 +137,15 @@ export class Client {
     this.#isClosed = true;
   }
 
+  paginate<T extends QueryValue>(
```
cleve-fauna, line 140: add documentation to this method.

src/values/set.ts
```diff
@@ -0,0 +1,181 @@
...
+  readonly after?: string;
+
+  constructor({ data, after }: { data: T[]; after?: string }) {
+    this.data = data;
+    this.after = after;
+  }
+}
+
+/**
+ * A un-materialize Set. Typically received when a materialized Set contains
```
cleve-fauna, line 26:
```suggestion
 * A un-materialized Set. Typically received when a materialized Set contains
```

src/values/set.ts
```diff
@@ -0,0 +1,181 @@
...
+/**
+ * A class to provide an iterable API for fetching multiple pages of data, given
+ * a Fauna Set
+ */
+export class SetIterator<T extends QueryValue>
+  implements AsyncGenerator<T[], void, unknown>
+{
+  readonly #generator: AsyncGenerator<T[], void, unknown>;
+
+  constructor(
```
cleve-fauna, line 51: add documentation.

src/values/set.ts
```diff
@@ -0,0 +1,181 @@
...
+    client: Client,
+    initial: Page<T> | EmbeddedSet | (() => Promise<T>)
+  ) {
+    if (initial instanceof Function) {
+      this.#generator = generateFromThunk(client, initial);
+    } else if (initial instanceof Page || initial instanceof EmbeddedSet) {
+      this.#generator = generatePages(client, initial);
+    } else {
+      throw new TypeError(
+        "Expected 'Pageable<QueryValue> | (() => Promise<T>)', but received " +
```
cleve-fauna, line 61: There's no such construct as `Pageable`. 

Rewrite your error message to refer to what actually exists (Page or EmbeddedSet)

If they share some interface I suppose you could have them both implement a "Pageable" interface you define as another alternative.

src/values/set.ts
```diff
@@ -0,0 +1,181 @@
...
+          // @ts-expect-error "Property 'constructor' does not exist on type 'never'."
+          // This is okay, we still want to catch weird inputs from JS
+          initial.constructor.name +
+          " " +
+          JSON.stringify(initial)
```
cleve-fauna, line 66: this seems like it could blow up.

I don't think you need to include that.

src/values/set.ts
```diff
@@ -0,0 +1,181 @@
...
+  constructor(after: string) {
+    this.after = after;
+  }
+}
+
+/**
+ * A class to provide an iterable API for fetching multiple pages of data, given
+ * a Fauna Set
+ */
+export class SetIterator<T extends QueryValue>
```
cleve-fauna, line 46: We should give users an optional page size control.

src/values/set.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+}
+
+export class FlattenedSetIterator<T extends QueryValue>
+  implements AsyncGenerator<T, void, unknown>
+{
+  readonly #generator: AsyncGenerator<T, void, unknown>;
+
+  constructor(setIterator: SetIterator<T>) {
+    async function* generateItems<T extends QueryValue>(
```
cleve-fauna, line 115: instead of rebuilding this function everytime why not move it out to the broader scope like you did the other generator functions?

cleve-fauna (commented):
Nice work Paul!

A few suggestions.

src/values/set.ts
ptpaterson: Great catch, I did have a type alias for a hot second there, but this needs fixed 👍🏻

src/values/set.ts
ptpaterson: In what way? Even variables with literal values have constructors

```javascript
const x = 5
console.log(x.constructor.name) // Number
```

I guess if you pass in an object that has a constructor property this would make a nonsense error. I'd rather take out the `.constructor.name` part and leave the error though.

Types will be enforced by typescript of course, but for JS it will be easy to pass in nonsense. I think this is one of the places that would be good to catch bad input early when otherwise behavior would be undefined.

cleve-fauna: sure i'm not saying remove the validation. I could just imagine a JS object without a contructor field.

src/values/set.ts
ptpaterson: Good idea! Current caveat is that the first page will always be up to 16. We can fix eventually by adding `.paginate(${size})` to the end of their query, in the case they use `client.paginate` or otherwise use `SetIterator.fromQuery`, but the `<set>.paginate` method doesn't yet return `@set`-tagged values.

ptpaterson: hmmmm. should we superficially just return groups of the specified size, regardless if the result is larger (should only be necessary for the first page)

src/values/set.ts
ptpaterson: I started there and then was trying to keep them organized together. But I suppose that's in consistent with the other functions. I'll move this out.

src/values/set.ts
ptpaterson: Fair enough. Adding in size is additive, so it's certainly safe to hold off if anything. I think it is also fair to require users to use `.paginate(number)` directly (once core makes that return the tagged value).

cleve-fauna: yeah i think if core handled it it would be a simpler API as we are just using core then and not doing any shimming.

src/values/set.ts
```diff
@@ -0,0 +1,250 @@
+import { Client } from "../client";
+import { Query, fql } from "../query-builder";
+import { QueryValue } from "../wire-protocol";
+
+/**
+ * A materialize view of a Set.
```
cleve-fauna, line 6:
```suggestion
 * A materialized view of a Set.
```

cleve-fauna (approved):

PR #144, opened 2023-05-05
Refactor QueryBuilder to remove dependency on encoder

ptpaterson:
Ticket(s): [FE-3179](https://faunadb.atlassian.net/browse/FE-3179)

## Problem
The Query class calls encoding, and the encoder cannot work with nested Query objects.

## Solution
Separate concerns of building queries from encoding them

## Result
The Query class is a container for query string fragments and arguments. The encoder does all of the work to encode arguments, including nested Query objects.

It no longer makes sense for the Query object to handle information about query options/headers, so cleaned up some functionality around query options and direct arguments.

## Out of scope
Handling Query objects within objects and arrays.

## Testing
Added a bunch of more tests. We were previously testing the query-builder with only the tagged format, so I added tests and expectations for the simple format as well.

Added a test for providing direct arguments, which was always possible if not prominently exposed.

----

[FE-3179]: https://faunadb.atlassian.net/browse/FE-3179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/wire-protocol.ts
+++ b/src/wire-protocol.ts
@@ -1,5 +1,5 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-import { fql } from "./query-builder";
+import { Query, fql } from "./query-builder";
 import {
   DateStub,
   Document,
@@ -16,7 +16,7 @@ import {
 /**
  * A request to make to Fauna.
  */
-export interface QueryRequest extends QueryRequestHeaders {
+export interface QueryRequest {
   /** The query */
   query: string | QueryInterpolation;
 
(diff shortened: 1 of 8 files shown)


src/wire-protocol.ts
```diff
@@ -16,7 +16,7 @@ import {
 /**
  * A request to make to Fauna.
  */
-export interface QueryRequest extends QueryRequestHeaders {
```
ptpaterson, line 19: @cleve-fauna We have already changed the signature of Client.query to accept a `Query` (previously `QueryBuilder`) and not `Query | QueryRequest`. That makes `QueryRequest` specifically just the thing we send over the wire, and it doesn't make sense for it to also have the headers fields. (We were probably sending the options over the wire for a time, weren't we 🤔 )

It also mean that the note for `Client.query` that "you can embed header fields in this object; if you do that there's no need to pass the headers parameter." is no longer accurate... I didn't catch that before now.

As a side-effect, type enforcement for `QueryRequest` and `QueryRequestOptions` in methods like `Client.query` and `Client.#setHeaders` is more accurate.

PR #145, opened 2023-05-05
Implement http2SessionIdleTime and ClientTimeout

ptpaterson:
Ticket(s): [FE-3508](https://faunadb.atlassian.net/browse/FE-3508)

## Problem
Users need to be able to configure http2SessionIdleTime and ClientTimeout

## Solution
Convert `HTTPClient` interface to an abstract class with configuration properties: `client_timeout_ms` and `http2_sessions_idle_ms`. `FetchClient` and `NodeHTTP2Client` implement this class and can accept configuration options.

## Result
WIP:

- [ x ] `client_timeout_ms` for fetch client
- [ ] `client_timeout_ms` for node http client
- [ ] `http2_sessions_idle_ms` for node http client
- [ ] wire the options up to the client configuration
- [ ] validate that if `client_timeout_ms` is provided, that `query_timeout_ms` is also provided and `client_timeout_ms > query_timeout_ms + n`. Also validate this relationship if `query_timeout_ms` is provided as a per-query option.

## Out of scope
keepalive for fetch client

## Testing
Adding tests

----

### DIFF
--- a/src/http-client/fetch-client.ts
+++ b/src/http-client/fetch-client.ts
@@ -2,12 +2,21 @@
 /// <reference lib="dom" />
 
 import { NetworkError } from "../errors";
-import { HTTPClient, HTTPRequest, HTTPResponse } from "./index";
+import {
+  HTTPClient,
+  HTTPClientOptions,
+  HTTPRequest,
+  HTTPResponse,
+} from "./http-client";
 
 /**
  * An implementation for {@link HTTPClient} that uses the native fetch API
  */
-export class FetchClient implements HTTPClient {
+export class FetchClient extends HTTPClient {
+  constructor(options: HTTPClientOptions) {
+    super(options);
+  }
+
   /** {@inheritDoc HTTPClient.request} */
   async request({
     data,
@@ -17,10 +26,19 @@ export class FetchClient implements HTTPClient {
   }: HTTPRequest): Promise<HTTPResponse> {
     // TODO: handle client timeouts with AbortController. Emit NetworkError if so.
 
+    let controller: AbortController;
+    let signal: AbortSignal | undefined = undefined;
+    if (this.client_timeout_ms !== undefined) {
+      controller = new AbortController();
+      signal = controller.signal;
+      setTimeout(() => controller.abort(), this.client_timeout_ms);
+    }
+
     const response = await fetch(url, {
       method,
       headers: { ...requestHeaders, "Content-Type": "application/json" },
       body: JSON.stringify(data),
+      signal,
     }).catch((error) => {
       throw new NetworkError("The network connection encountered a problem.", {
         cause: error,
--- a/src/http-client/http-client.ts
+++ b/src/http-client/http-client.ts
@@ -1,0 +1,84 @@
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import type { Client } from "../client";
+import { QueryRequest } from "../wire-protocol";
+
+/**
+ * An object representing an http request.
+ * The {@link Client} provides this to the {@link HTTPClient} implementation.
+ */
+export type HTTPRequest = {
+  data: QueryRequest;
+  headers: Record<string, string | undefined>;
+  method: "POST";
+  url: string;
+};
+
+/**
+ * An object representing an http request.
+ * It is returned to, and handled by, the {@link Client}.
+ */
+export type HTTPResponse = {
+  body: string;
+  headers: Record<string, string | string[] | undefined>;
+  status: number;
+};
+
+export type HTTPClientOptions = {
+  /**
+   * Time in milliseconds at which the client will abort a request if it has not
+   * received a response
+   */
+  client_timeout_ms?: number;
+
+  /**
+   * Time in milliseconds the client will keep an HTTP2 session open after all
+   * requests are completed. Only necessary for HTTP2 implementations.
+   */
+  http2_sessions_idle_ms?: number;
+};
+
+/**
+ * An interface to provide implementation-specific, asyncronous http calls.
+ * This driver provides default implementations for common environments. Users
+ * can configure the {@link Client} to use custom implementations if desired.
+ */
+export abstract class HTTPClient {
+  /**
+   * Time in milliseconds at which the client will abort a request if it has not
+   * received a response
+   */
+  client_timeout_ms?: number;
+
+  /**
+   * Time in milliseconds the client will keep an HTTP2 session open after all
+   * requests are completed. Only necessary for HTTP2 implementations.
+   */
+  http2_sessions_idle_ms?: number;
+
+  constructor({
+    client_timeout_ms,
+    http2_sessions_idle_ms,
+  }: HTTPClientOptions) {
+    this.client_timeout_ms = client_timeout_ms;
+    this.http2_sessions_idle_ms = http2_sessions_idle_ms;
+  }
+
+  /**
+   * Makes an HTTP request and returns the response
+   * @param req - an {@link HTTPRequest}
+   * @returns A Promise&lt;{@link HTTPResponse}&gt;
+   */
+  abstract request(req: HTTPRequest): Promise<HTTPResponse>;
+
+  /**
+   * Flags the calling {@link Client} as no longer
+   * referencing this HTTPClient. Once no {@link Client} instances reference this HTTPClient
+   * the underlying resources will be closed.
+   * It is expected that calls to this method are _only_ made by a {@link Client}
+   * instantiation. The behavior of direct calls is undefined.
+   * @remarks
+   * For some HTTPClients, such as the {@link FetchClient}, this method
+   * is a no-op as there is no shared resource to close.
+   */
+  abstract close(): void;
+}
--- a/src/http-client/index.ts
+++ b/src/http-client/index.ts
@@ -1,68 +1,19 @@
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-import type { Client } from "../client";
-import { QueryRequest } from "../wire-protocol";
 import { FetchClient } from "./fetch-client";
 import { NodeHTTP2Client } from "./node-http2-client";
+import {
+  HTTPClient,
+  HTTPClientOptions,
+  HTTPRequest,
+  HTTPResponse,
+} from "./http-client";
 
-export { FetchClient } from "./fetch-client";
-export { NodeHTTP2Client } from "./node-http2-client";
-
-/**
- * An object representing an http request.
- * The {@link Client} provides this to the {@link HTTPClient} implementation.
- */
-export type HTTPRequest = {
-  data: QueryRequest;
-  headers: Record<string, string | undefined>;
-  method: "POST";
-  url: string;
-};
-
-/**
- * An object representing an http request.
- * It is returned to, and handled by, the {@link Client}.
- */
-export type HTTPResponse = {
-  body: string;
-  headers: Record<string, string | string[] | undefined>;
-  status: number;
-};
-
-/**
- * An interface to provide implementation-specific, asyncronous http calls.
- * This driver provides default implementations for common environments. Users
- * can configure the {@link Client} to use custom implementations if desired.
- */
-export interface HTTPClient {
-  /**
-   * Makes an HTTP request and returns the response
-   * @param req - an {@link HTTPRequest}
-   * @returns A Promise&lt;{@link HTTPResponse}&gt;
-   */
-  request(req: HTTPRequest): Promise<HTTPResponse>;
-
-  /**
-   * Flags the calling {@link Client} as no longer
-   * referencing this HTTPClient. Once no {@link Client} instances reference this HTTPClient
-   * the underlying resources will be closed.
-   * It is expected that calls to this method are _only_ made by a {@link Client}
-   * instantiation. The behavior of direct calls is undefined.
-   * @remarks
-   * For some HTTPClients, such as the {@link FetchClient}, this method
-   * is a no-op as there is no shared resource to close.
-   */
-  close(): void;
-}
-
-export const getDefaultHTTPClient = () =>
-  isNode() ? NodeHTTP2Client.getClient() : new FetchClient();
-
-// utility functions
+export const getDefaultHTTPClient = (): HTTPClient =>
+  isNode() ? NodeHTTP2Client.getClient() : new FetchClient({});
 
 export const isHTTPResponse = (res: any): res is HTTPResponse =>
   res instanceof Object && "body" in res && "headers" in res && "status" in res;
 
-export function isNode() {
+const isNode = () => {
   if (typeof process !== "undefined" && process.release?.name === "node") {
     try {
       require("node:http2");
@@ -72,4 +23,13 @@ export function isNode() {
     }
   }
   return false;
-}
+};
+
+export {
+  FetchClient,
+  NodeHTTP2Client,
+  HTTPClient,
+  HTTPClientOptions,
+  HTTPRequest,
+  HTTPResponse,
+};
--- a/src/http-client/node-http2-client.ts
+++ b/src/http-client/node-http2-client.ts
@@ -4,14 +4,19 @@ try {
 } catch (_) {
   http2 = undefined;
 }
-import { HTTPClient, HTTPRequest, HTTPResponse } from "./index";
+import {
+  HTTPClient,
+  HTTPClientOptions,
+  HTTPRequest,
+  HTTPResponse,
+} from "./http-client";
 import { NetworkError } from "../errors";
 import { QueryRequest } from "../wire-protocol";
 
 /**
  * An implementation for {@link HTTPClient} that uses the node http package
  */
-export class NodeHTTP2Client implements HTTPClient {
+export class NodeHTTP2Client extends HTTPClient {
   static #client: NodeHTTP2Client | null = null;
 
   #sessionMap: Map<string, SessionWrapper> = new Map();
@@ -22,8 +27,9 @@ export class NodeHTTP2Client implements HTTPClient {
    * @remarks Private constructor means you must instantiate with
    * {@link NodeHTTP2Client.getClient}
    */
-  // eslint-disable-next-line @typescript-eslint/no-empty-function
-  private constructor() {}
+  private constructor(options: HTTPClientOptions) {
+    super(options);
+  }
 
   static getClient() {
     if (http2 === undefined) {
@@ -31,7 +37,7 @@ export class NodeHTTP2Client implements HTTPClient {
     }
 
     if (this.#client === null) {
-      this.#client = new NodeHTTP2Client();
+      this.#client = new NodeHTTP2Client({});
     }
 
     this.#client.#numberOfUsers++;
--- a/__tests__/unit/fetch-client.test.ts
+++ b/__tests__/unit/fetch-client.test.ts
@@ -32,7 +32,7 @@ const dummyStats = {
 
 describe("fetch client", () => {
   beforeAll(() => {
-    fetchClient = new FetchClient();
+    fetchClient = new FetchClient({ client_timeout_ms: 65_000 });
   });
 
   afterAll(() => {
@@ -112,6 +112,26 @@ describe("fetch client", () => {
     fetchMock.mockRejectOnce(new Error("oops"));
     try {
       await fetchClient.request(dummyRequest);
+    } catch (e) {
+      if (e instanceof NetworkError) {
+        expect(e.message).toEqual(
+          "The network connection encountered a problem."
+        );
+        expect(e.cause).toBeDefined();
+      }
+    }
+  });
+
+  it("returns a NetworkError if client timeout causes an abort", async () => {
+    expect.assertions(2);
+    fetchMock.mockResponseOnce(
+      () =>
+        new Promise((resolve) => setTimeout(() => resolve({ body: "" }), 100))
+    );
+    try {
+      const badClient = new FetchClient({ client_timeout_ms: 10 });
+
+      await badClient.request(dummyRequest);
     } catch (e) {
       if (e instanceof NetworkError) {
         expect(e.message).toEqual(
--- a/__tests__/unit/query.test.ts
+++ b/__tests__/unit/query.test.ts
@@ -23,7 +23,7 @@ const client = getClient(
     query_timeout_ms: 60,
   },
   // use the FetchClient implementation, so we can mock requests
-  new FetchClient()
+  new FetchClient({})
 );
 
 describe("query", () => {


ptpaterson:
Bah. Instance properties doesn't work well when we're using a singleton! The properties are set the first time and then all requests use that client.

I will update so that we have to pass down the configuration options on each request. We can go back to the simpler interface, rather than the abstract class.

src/client.ts
```diff
@@ -425,14 +432,49 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+
+  #validateRequest(_: Query, queryOptions?: QueryRequestHeaders) {
+    if (
+      queryOptions?.query_timeout_ms &&
+      this.#clientConfiguration.client_timeout_ms &&
+      this.#clientConfiguration.client_timeout_ms <
+        queryOptions.query_timeout_ms + CLIENT_TIMEOUT_BUFFER
+    ) {
+      throw new RangeError(
+        `'query_timeout_ms' is more than maximum allowed setting. 'clien_timeout_ms' must be greater than or equal to 'query_timeout_ms' + ${CLIENT_TIMEOUT_BUFFER} ms`
```
cleve-fauna, line 464: I suppose another option woudl be to let uers specify only a queryTImeout and a clientTimeoutBuffer 🤔 then they don't have to maintain two timeout settings.

cleve-fauna (commented):
Some early feedback for you

src/client.ts
```diff
@@ -425,14 +432,49 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
   }
+
+  #validateConfiguration() {
+    const config = this.#clientConfiguration;
+    if (config.client_timeout_ms) {
+      if (!config.query_timeout_ms) {
+        throw new TypeError(
+          `Expected number, but received ${config.query_timeout_ms}. 'query_timeout_ms' must be configured if 'client_timeout_ms' is configured.`
+        );
+      }
```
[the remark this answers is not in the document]
ptpaterson: Default is "there is no query_timeout_ms". Our Jira says that default is to not provide a query timeout, so that core can handle the default.

This goes away though if we have change to the concept of a configured buffer, rather than a configured absolute value.

src/client-configuration.ts
```diff
@@ -25,28 +26,47 @@ export interface ClientConfiguration {
...
+   */
+  http2_session_idle_ms?: number;
+
+  // Default query options
+
   /**
    * The timeout of each query, in milliseconds. This controls the maximum amount of
    * time Fauna will execute your query before marking it failed.
    * Default is undefined which let's Fauna determine the query timeout to apply. This
    * is recommended for most queries.
```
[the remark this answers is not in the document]
ptpaterson: This is configurable by core, I don't think we want to talk about what the default value is. The driver provides no value by default, so falls back on core.

__tests__/functional/client-configuration.test.ts
```diff
@@ -147,4 +147,22 @@ an environmental variable named FAUNA_SECRET or pass it to the Client constructo
...
   });
+
+  it("throws an error if 'query_timeout_ms' is not configured when 'client_timeout_ms' is.", async () => {
+    expect.assertions(1);
+    try {
+      getClient({ client_timeout_ms: 65_000 });
+    } catch (e: any) {
+      expect(e).toBeInstanceOf(TypeError);
+    }
+  });
```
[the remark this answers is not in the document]
ptpaterson: Changed to the buffer api. We will now only validate and throw when configuring the client with buffer < 0.

src/client-configuration.ts
```diff
@@ -25,28 +26,47 @@ export interface ClientConfiguration {
...
    * @see https://docs.fauna.com/fauna/current/security/keys
    */
   secret: string;
+
+  /**
+   * Time in milliseconds at which the client will abort a request if it has not
+   * received a response
+   */
+  client_timeout_ms?: number;
+
```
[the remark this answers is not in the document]
ptpaterson: Buffer defaults to 500 ms (at time of writing), and I added a note about it.

src/http-client/index.ts
```diff
@@ -1,68 +1,14 @@
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
ptpaterson, line 1: I ran into circular dependency issues, so that's why I broke out the type definitions to a separate module.

src/util.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Used for functions that take an options objects. Fills in defaults for
+ * options not provided.
+ *
+ * @internal
+ */
+export const applyDefaults = <T extends Record<string, any>>(
```
ptpaterson, line 7: I was fighting a bug for a while where the "provided" value had field present but set to undefined. This caused the default value to be overridden with undefined. This utility is safer than Object.assign or spread operators for applying defaults in case the provided value contains any undefined fields.

__tests__/functional/client-configuration.test.ts
```diff
@@ -147,4 +147,13 @@ an environmental variable named FAUNA_SECRET or pass it to the Client constructo
     );
     await client2.query<number>(fql`"taco".length`);
   });
+
+  it("throws a RangeError if 'client_timeout_buffer_ms' is less than zero", async () => {
```
cleve-fauna, line 151: we should force it to be > 0.

src/client-configuration.ts
```diff
@@ -25,28 +36,42 @@ export interface ClientConfiguration {
    * Fauna as a "Date", but will instead be treated as a string.
    */
   format: ValueFormat;
+
+  /**
+   * Time in milliseconds the client will keep an HTTP2 session open after all
+   * requests are completed. Only necessary for HTTP2 implementations. The
+   * default is 500 ms.
```
cleve-fauna, line 43:
```suggestion
   * requests are completed. The default is 500 ms.
```

src/util.ts
cleve-fauna: Sounds a like a user-level bug.

Basically this function ingores undefined values.

There may be a scenario in the future where a user wants to override some default setting to undefined  - imagine if we had a new client config control that was on by default but could be turned on by setting to undefined.

IMO we should keep using the spread operators. That way the interface is simple - you get what you ask for. If you ask for undefined on a value that's what you get.

If the bug was hard to track down we can add some client config validation to catch those immediately and give a clean error to the user.

src/http-client/node-http2-client.ts
```diff
@@ -149,7 +140,8 @@ class SessionWrapper {
     data: requestData,
     headers: requestHeaders,
     method,
-  }: SessionRequestOptions): Promise<HTTPResponse> {
+    client_timeout_ms,
```
cleve-fauna, line 143: let's make life easy for ourselves and use the buffer here.

So the buffer is always controlled in one palce.

src/http-client/node-http2-client.ts
```diff
@@ -188,6 +180,14 @@ class SessionWrapper {
           .on("error", (error: any) => rejectPromise(error))
           .on("response", onResponse);
         req.write(JSON.stringify(requestData), "utf8");
+
+        // req.setTimeout must be called before req.end()
+        if (client_timeout_ms !== undefined) {
```
cleve-fauna, line 185: this should never be undefined.

We should always have some kind of client timeout (though it wll be based on the query timeout + the buffer)

src/http-client/http-client.ts
```diff
@@ -0,0 +1,69 @@
...
+  method: "POST";
+
+  /** URL for the request */
+  url: string;
+
+  /**
+   * The timeout of each query, in milliseconds. This controls the maximum amount of
+   * time Fauna will execute your query before marking it failed.
+   */
+  client_timeout_ms?: number;
```
cleve-fauna, line 32: this is confusing.

This is not really controlling a query timeout - its the HTTP client timeout.

Let's set this as the buffer as well.

We can pass down a query_timeout_ms in this type as well.

That way the client timeout can use the simple query_timeout_ms + client_timeout_ms math.

If query_timeout is undefined then we know to use core's default as the query_timeout_ms basis.

src/http-client/fetch-client.ts
```diff
@@ -14,13 +14,23 @@ export class FetchClient implements HTTPClient {
     headers: requestHeaders,
     method,
     url,
+    client_timeout_ms,
   }: HTTPRequest): Promise<HTTPResponse> {
     // TODO: handle client timeouts with AbortController. Emit NetworkError if so.
 
+    let controller: AbortController;
+    let signal: AbortSignal | undefined = undefined;
+    if (client_timeout_ms !== undefined) {
```
cleve-fauna, line 23: this concept should never be undefined.

ALso, let's use the buffer here added to the query timeout.

src/client.ts
```diff
@@ -425,11 +445,20 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
       headerObject["x-last-txn-ts"] = this.#lastTxnTs;
     }
   }
+
+  #validateConfiguration() {
+    const config = this.#clientConfiguration;
+    if (config.client_timeout_buffer_ms < 0) {
```
cleve-fauna, line 451:
```suggestion
    if (config.client_timeout_buffer_ms <= 0) {
```

src/http-client/http-client.ts
cleve-fauna: alternative is to infer the client timeout from defaults in the client.ts and never pass undefined here. and have the buffer be a client only concern. that might be cleaner

src/http-client/node-http2-client.ts
cleve-fauna: alternative is to infer the client timeout from defaults in the client.ts and never pass undefined here. and have the buffer be a client only concern. that might be cleaner

src/util.ts
ptpaterson: There is MUCH to be said about keeping it simple! And I sometimes fall in to a trap of making things more complicated than they need to be, so this is always good to be checked on 🙂. Here's some background on this one:

I created the bug within the driver by defining one of the timeouts as possibly undefined. That explicit undefined value was passed down through the HTTPRequest. At the time, the default was being applied/enforced in the HTTPClient and was getting overridden by that explicit undefined. I noticed the potential for this issue at client configuration, too, so created the reusable util based on the [v4 driver](https://github.com/fauna/faunadb-js/blob/3940ad9e71907b7795fd975ae16c9392a679b053/src/_util.js#L348-L365). (We have the benefit of ES6 and typescript to simplify the method and enforce mandatory options without setting undefined in the default)

This is not relevant to the HTTPCient anymore (timeouts and session idle time are defined before passing to the http client), but I think it is still important to apply defaults this way. That is, important to not let explicit undefined options override the defaults. We rely on the defaults being present, and now we have more of them.

```typescript
export const DEFAULT_CLIENT_CONFIG: Omit<ClientConfiguration, "secret"> = {
  client_timeout_buffer_ms: 500,
  endpoint: endpoints.default,
  format: "tagged",
  http2_session_idle_ms: 500,
  max_conns: 10,
  query_timeout_ms: 5000
};
```

We added `client_timeout_buffer_ms`, `http2_session_idle_ms`, and `query_timeout_ms` to mandatory client configuration. The driver assumes these values are defined, so will blow up they are set to undefined.

src/client-configuration.ts
```diff
@@ -4,10 +4,21 @@ import { ValueFormat } from "./wire-protocol";
  * Configuration for a client.
  */
 export interface ClientConfiguration {
+  /**
+   * Time in milliseconds beyond {@link ClientConfiguration.query_timeout_ms} at
+   * which the client will abort a request if it has not received a response.
+   * The default is 500 ms, which should account for network latency for most
+   * clients. The value must be greater than zero. The closer to zero the value
+   * is, the more likely the client is to abort the request before the server
+   * can report a legitimate timeout error.
```
cleve-fauna, line 13:
```suggestion
   * can report a legitimate response or error.
```

src/client-configuration.ts
```diff
@@ -25,28 +36,41 @@ export interface ClientConfiguration {
...
    */
   secret: string;
+
+  // Query options
+
   /**
    * The timeout of each query, in milliseconds. This controls the maximum amount of
    * time Fauna will execute your query before marking it failed.
    * Default is undefined which let's Fauna determine the query timeout to apply. This
    * is recommended for most queries.
```
cleve-fauna, line 63:
```suggestion
   * is recommended for most queries. The default is 5000 ms.
```

src/client.ts
```diff
@@ -425,11 +449,20 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
       headerObject["x-last-txn-ts"] = this.#lastTxnTs;
     }
   }
+
+  #validateConfiguration() {
```
cleve-fauna, line 453: query timeout must also be > 0

src/client.ts
```diff
@@ -79,15 +80,21 @@ export class Client {
     clientConfiguration?: Partial<ClientConfiguration>,
     httpClient?: HTTPClient
   ) {
-    this.#clientConfiguration = {
-      ...defaultClientConfiguration,
-      ...clientConfiguration,
-      secret: this.#getSecret(clientConfiguration),
-    };
+    this.#clientConfiguration = applyDefaults(
```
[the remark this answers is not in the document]
ptpaterson: There is no way to be selective about which fields are permitted to be explicitly provided as undefined or not. JS is just crappy that way. So, since we know that the driver will behave incorrectly when some options are undefined, I can instead enumerate each one and individually ignore undefined values. For client configuration, at this time, that will amount to the exact same behavior (except for max_conns, because that's not connected to anything yet).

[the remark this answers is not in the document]
cleve-fauna: yeah i understand that, but we could expand validateCOnfiguration to catch those cases and keep the interface very simple to understand.

Treating undefined special makes the interface more complex

src/http-client/http-client.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  /** Headers in object format */
+  headers: Record<string, string | undefined>;
+
+  /**
+   * Time in milliseconds the client will keep an HTTP2 session open after all
+   * requests are completed. Only necessary for HTTP2 implementations.
+   */
+  http2_sessions_idle_ms: number;
+
```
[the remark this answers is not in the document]
cleve-fauna: Yes, I can see that. but this is a bit of putting a square peg in a round hole. See the comment in node-http2-client.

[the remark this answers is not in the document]
ptpaterson: I also just saw your other comment about changing session api. That may change things.

src/client.ts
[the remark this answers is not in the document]
ptpaterson: Updated to throw error when the required options are undefined. We will no longer apply the defaults in that case.

src/http-client/node-http2-client.ts
```diff
@@ -74,8 +73,8 @@ export class NodeHTTP2Client implements HTTPClient {
     return this.#numberOfUsers === 0;
   }
 
-  #getSession(url: string): SessionWrapper {
-    const sessionKey = url; // WIP: need to account for streaming
+  #getSession(request: HTTPRequest): SessionWrapper {
```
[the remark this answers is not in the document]
ptpaterson: We also need different sessions (eventually) for streaming, request of which are sent to a different path. So I am trying to think through how we won't need to open a connection immediately. Every request still needs to check if the session is present and if not, then create a new one.

I also prefer option 1 over 2.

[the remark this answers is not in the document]
cleve-fauna: We are not building streaming right now so let's not concern ourselves with it.

We might need to rethink the streaming product to make it more appealing.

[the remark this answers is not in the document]
ptpaterson: Tell me if I am on the right track here:

Upon client configuration, we should know what the session hash key(s) are, and will _always_ be for each request. So we can calculate that in advance and save the session configuration. Then on each request, we can check if the session is still valid and if not create a new one with the existing configuration (url, method, idle time).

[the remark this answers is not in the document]
cleve-fauna: Let's just huddle it will be easier.

src/http-client/node-http2-client.ts
```diff
@@ -55,143 +120,75 @@ export class NodeHTTP2Client implements HTTPClient {
...
   }
 }
+
+/**
+ * Put the session in an object, so we can update the session without deleting
+ * and resetting what is in the Map
+ */
+type SessionWrapper = {
+  session: ClientHttp2Session;
+};
```
ptpaterson, line 194: Now that I have everything working, I don't think this is necessary anymore. But when I naively unwrapped everything to just put a ClientHttp2Session in the static session map, I got unexpected errors from the test to check if sessions are closed. I will need more time to figure this one out we want to drop it. It's not too intrusive as is, though.

__tests__/unit/client.test.ts
```diff
@@ -33,26 +37,26 @@ describe("Client", () => {
...
     );
+    expect(nodeClient.isClosed()).toBe(false);
     const clientThree = new Client(
       getDefaultSecretAndEndpoint(),
-      NodeHTTP2Client.getClient()
+      new NodeHTTP2Client(getDefaultHTTPClientOptions())
     );
     clientTwo.close();
-    expect(nodeClient.isClosed()).toBe(false);
+    expect(nodeClient.isClosed()).toBe(true);
```
ptpaterson, line 59: There isn't ref counting anymore. We can reimplement it, but I took it out because I was struggling to make it correct.

I ran with this because I think it is safe (if not most efficient) for this iteration as clients with different connection properties (url and session idle time) will use completely different sessions. Any client that has finds itself with a close session will create a new one. I also think there were issues with the ref counting before, for example

- all sessions get closed. numberOfUsers == 0.
- the first client to start back up gets the session from the map, sees it's closed, replaces with new session. numberOfUsers == 0
- the next client to make a requests get the latest session,m which is open.
- one of the clients calls `closed` numberOfUsers is 0 so session.close is called

src/http-client/node-http2-client.ts
```diff
@@ -4,47 +4,112 @@ try {
 } catch (_) {
   http2 = undefined;
 }
-import { HTTPClient, HTTPRequest, HTTPResponse } from "./index";
+// import http2 from "node:http2";
```
cleve-fauna, line 7:
```suggestion
```

src/http-client/node-http2-client.ts
```diff
@@ -4,47 +4,112 @@ try {
...
+type IncomingHttpHeaders = any;
+type IncomingHttpStatusHeader = any;
+type OutgoingHttpHeaders = any;
+
+// import http2 from "node:http2";
+// type ClientHttp2Session = http2.ClientHttp2Session
+// type ClientHttp2Stream = http2.ClientHttp2Stream
+// type IncomingHttpHeaders = http2.IncomingHttpHeaders
+// type IncomingHttpStatusHeader = http2.IncomingHttpStatusHeader
+// type OutgoingHttpHeaders = http2.OutgoingHttpHeaders
```
cleve-fauna, line 28:
```suggestion
```

cleve-fauna (commented):
Hey Paul we can keep the ref counting and the implementation of `NodeHTTP2Client` largely unchagnged if we store a `Map<string, NodeHTTP2Client>` statically rather than a `Map<string, SessionWrapper>`.

So when getClient gets called it looks in the map and calls some type of construct on NodeHTTP2Client if the hash key doesn't exist - otherwise it uses the one matching the hash key. In both cases it increments the ref count.

__tests__/unit/client.test.ts
```diff
@@ -33,26 +37,26 @@ describe("Client", () => {
   });
 
   it("Allows for creation and usage of a new client after first client closed.", async () => {
-    const nodeClient = NodeHTTP2Client.getClient();
+    const nodeClient = new NodeHTTP2Client(getDefaultHTTPClientOptions());
     const clientOne = new Client(getDefaultSecretAndEndpoint(), nodeClient);
-    expect(nodeClient.isClosed()).toBe(false);
     await clientOne.query(fql`"Hello World"`);
+    expect(nodeClient.isClosed()).toBe(false);
```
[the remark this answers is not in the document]
ptpaterson: Adding reference counting back, but this is still true. Sessions appear effectively closed until you make a request.

src/http-client/node-http2-client.ts
ptpaterson: I'm using this to hold onto references, and renamed to SessionRC in the latest update.

cleve-fauna:
Discussed offline trying to ship this for now and then following on with a patch to make the code more space efficient.

src/http-client/node-http2-client.ts
```diff
@@ -4,47 +4,106 @@ try {
...
+type ClientHttp2Session = any;
+type ClientHttp2Stream = any;
+type IncomingHttpHeaders = any;
+type IncomingHttpStatusHeader = any;
+type OutgoingHttpHeaders = any;
 
 /**
  * An implementation for {@link HTTPClient} that uses the node http package
  */
 export class NodeHTTP2Client implements HTTPClient {
```
cleve-fauna, line 25: as discussed offline, this implementation will store `O(n)` resources where `(n)` is the number of Fauna Client a user is using.

However, this PR captures a nice interface.

So we will ship this and follow on with a patch to move `O(n) -> O(s)` where `s` is the number different HttpOptions used by clients, and thus how many HTTP2 Sesions we must maintain - for 99% of users `s = 1`.

src/http-client/node-http2-client.ts
```diff
@@ -4,47 +4,106 @@ try {
...
+              body: responseData,
+              headers: http2ResponseHeaders,
+            });
+          });
+        };
+
+        try {
+          const httpRequestHeaders: OutgoingHttpHeaders = {
+            ...requestHeaders,
+            [http2.constants.HTTP2_HEADER_PATH]: "/query/1",
```
cleve-fauna, line 78: make this part of HttpOptions

src/http-client/node-http2-client.ts
```diff
@@ -53,145 +112,95 @@ export class NodeHTTP2Client implements HTTPClient {
...
+ */
+type SessionRC = {
+  /** A session to be shared among http clients */
+  session: ClientHttp2Session | null;
+  /**
+   * A Setof client references. We will only close the session when there are no
+   * references left. This cannot be a WeakSet, because WeakSets are not
+   * enumerable (can't check if they are empty).
+   */
+  refs: Set<NodeHTTP2Client>;
```
cleve-fauna, line 205: as discussed offline we can move back to user a simple count in a patch to save space.

src/client.ts
```diff
@@ -80,16 +78,18 @@ export class Client {
...
     this.#clientConfiguration = {
-      ...defaultClientConfiguration,
+      ...DEFAULT_CLIENT_CONFIG,
       ...clientConfiguration,
       secret: this.#getSecret(clientConfiguration),
     };
-    this.#url = new URL(
-      "/query/1",
-      this.clientConfiguration.endpoint
-    ).toString();
```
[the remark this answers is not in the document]
ptpaterson: This is added as the "path" for http2. I patched the FetchClient to make sure it's added there, too.

__tests__/integration/query.test.ts
```diff
@@ -337,6 +365,41 @@ describe("query", () => {
...
+    await new Promise((resolve) => setTimeout(resolve, 600));
+
+    // clients should be closed
+    expect(httpClient1.isClosed()).toBe(true);
+    expect(httpClient2.isClosed()).toBe(true);
+    expect(httpClient3.isClosed()).toBe(true);
+  });
+
+  it("can be called after session idle timeout", async () => {
+    const client = getClient({ http2_session_idle_ms: 500 });
```
cleve-fauna, line 392:
```suggestion
    const client = getClient({ http2_session_idle_ms: 5 });
```

500 is bigger than 100 😆

src/http-client/index.ts
```diff
@@ -1,68 +1,19 @@
...
-
-// utility functions
+export const getDefaultHTTPClient = (options: HTTPClientOptions): HTTPClient =>
+  isNode() ? new NodeHTTP2Client(options) : new FetchClient(options);
 
 export const isHTTPResponse = (res: any): res is HTTPResponse =>
   res instanceof Object && "body" in res && "headers" in res && "status" in res;
 
-export function isNode() {
+const isNode = () => {
```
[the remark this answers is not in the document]
cleve-fauna: 🧐 i say good sir :)

src/http-client/fetch-client.ts
```diff
@@ -2,25 +2,39 @@
...
+} from "./http-client";
 
 /**
  * An implementation for {@link HTTPClient} that uses the native fetch API
  */
 export class FetchClient implements HTTPClient {
+  #url: string;
+
+  constructor({ url }: HTTPClientOptions) {
+    this.#url = new URL("/query/1", url).toString();
```
cleve-fauna, line 19: not a big fan of hiding the URL we actually use in the HTTP client layer but we can revisit later.

src/http-client/node-http2-client.ts
cleve-fauna: we can revisit how to pass along a path / URL in the future.

src/http-client/node-http2-client.ts
```diff
@@ -4,47 +4,106 @@ try {
...
+        } catch (error) {
+          rejectPromise(error);
+        }
+      }
+    );
 
     try {
-      const result = await session.request(httpRequest);
-      return result;
+      return await requestPromise;
```
cleve-fauna, line 106: no need to return await.

```suggestion
      return requestPromise;
```

cleve-fauna (approved):

PR #147, opened 2023-05-10
Notify watchers when driver published

cleve-fauna:
## Problem
- stakeholders want to know when new driver versions are published
## Solution
- tag them in slack
## Result
- folks get notified

## Testing
- need to do a release and see if we ping OK
----


concourse/scripts/publish.sh
```diff
@@ -8,7 +8,7 @@ PACKAGE_VERSION=$(node -p -e "require('./package.json').version")
 NPM_LATEST_VERSION=$(npm view fauna version)
 echo "Current package version: $PACKAGE_VERSION"
 echo "Latest version in npm: $NPM_LATEST_VERSION"
-
+n
```
cleve-fauna, line 11:
```suggestion
```

henryfauna (approved):

PR #148, opened 2023-05-11
Update existing-collection.test.ts

cleve-fauna:
Ticket(s): FE-###

## Problem
- integ tests failing as FQL X now returns null docs not null on queries for byName

## Solution
- update our integ test code
## Result
- test no longer breaks

## Testing
- tests will not pass since we're running against public docker image
----


erickpintor (approved):
LGTM. Not sure if there are other tests doing the same thing, though.

cleve-fauna:
from what i understand this is no longer needed.

@erickpintor reopen if i'm wrong about that.

PR #157, opened 2023-05-15
Fingerprinting

ptpaterson:
Ticket(s): [FE-3011](https://faunadb.atlassian.net/browse/FE-3011)

## Problem
We don't send the driver environment details.

## Solution
Send them!

## Result
Gathers and formats information to send via the `X-Driver-Env` header.

## Testing
Test added to make sure the header exists.
Integration tests will make sure this works in other environments besides Node.

----

[FE-3011]: https://faunadb.atlassian.net/browse/FE-3011?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "@tsconfig/node18": "^1.0.1",
     "@types/jest": "^29.4.0",
     "@types/node": "^18.13.0",
+    "@types/serviceworker": "^0.0.67",
     "@typescript-eslint/eslint-plugin": "^5.52.0",
     "@typescript-eslint/parser": "^5.52.0",
     "esbuild": "^0.15.12",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "fauna",
-  "version": "0.7.1",
+  "version": "0.7.2",
   "description": "A driver to query Fauna databases in browsers, Node.js, and other Javascript runtimes",
   "homepage": "https://fauna.com",
   "bugs": {
--- a/src/client.ts
+++ b/src/client.ts
@@ -49,6 +49,9 @@ export const DEFAULT_CLIENT_CONFIG: Omit<ClientConfiguration, "secret"> = {
  * Client for calling Fauna.
  */
 export class Client {
+  /** A static copy of the driver env header to send with each request */
+  static readonly #driverEnvHeader = getDriverEnv();
+
   /** The {@link ClientConfiguration} */
   readonly #clientConfiguration: ClientConfiguration;
   /** The underlying {@link HTTPClient} client. */
@@ -441,7 +444,7 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
       headerObject["x-last-txn-ts"] = this.#lastTxnTs;
     }
 
-    headerObject["X-Driver-Env"] = getDriverEnv();
+    headerObject["x-driver-env"] = Client.#driverEnvHeader;
   }
 
   #validateConfiguration() {
--- a/src/client.ts
+++ b/src/client.ts
@@ -17,7 +17,15 @@ import {
   ContendedTransactionError,
   InvalidRequestError,
 } from "./errors";
+import {
+  getDefaultHTTPClient,
+  isHTTPResponse,
+  type HTTPClient,
+} from "./http-client";
 import { Query } from "./query-builder";
+import { TaggedTypeFormat } from "./tagged-type";
+import { getDriverEnv } from "./util/environment";
+import { EmbeddedSet, Page, SetIterator } from "./values";
 import {
   isQueryFailure,
   isQuerySuccess,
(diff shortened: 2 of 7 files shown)


tsconfig.json
```diff
@@ -6,7 +6,8 @@
     "outDir": "./dist",
     "declaration": true,
     "declarationDir": "./dist",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "resolveJsonModule": true
```
cleve-fauna, line 10: why is this added?

src/util/environment.ts
```diff
@@ -0,0 +1,329 @@
...
+  } catch (_) {
+    // okay if this fails
+  }
+
+  if (isNode()) {
+    driverEnv.runtime = ["nodejs", process.version].join("-");
+    driverEnv.env = getNodeRuntimeEnv();
+    driverEnv.os = [os.platform(), os.release()].join("-");
+  } else if (isServiceWorker) {
+    driverEnv.runtime = "Service Worker";
```
pnwpedro, line 51: Which cases does this condition cover, and are we sure that os won't be available/useful, or would it be better to try to obtain them?

ptpaterson: I don't think the executing context (global scope) for a service worker has anything else useful in it. I think it's just the [ServiceWorkerGlobalScope](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope).

ptpaterson: I see [ServiceWorkerGlobalScope.navigator](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/navigator), though. We should be able to get the browser info from that.

I'll work on that 👍🏻

ptpaterson: Cool! updated so ServiceWorker should send the same fields as a browser.

src/client.ts
```diff
@@ -439,6 +440,8 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
     ) {
       headerObject["x-last-txn-ts"] = this.#lastTxnTs;
     }
+
+    headerObject["X-Driver-Env"] = getDriverEnv();
```
cleve-fauna, line 444: let's cache this once at construction time so we don't have to farm this on every request.

src/util/environment.ts
```diff
@@ -0,0 +1,332 @@
...
+  if (typeof process !== "undefined" && process.release?.name === "node") {
+    try {
+      require("node:http2");
+      return true;
+    } catch (_) {
+      return false;
+    }
+  }
+  return false;
+};
```
cleve-fauna, line 24: this does not tell us if its node - but if its node with HTTP2.

We should use slightly different algorithms for fingerprinting vs determining a default client.

This one can be used for the default client - and can stay where it was (perhaps with a different name like useNodeHttp2) and a new one can just determine if its node or not.

src/util/environment.ts
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * Get OS details for the browser
+ */
+const getBrowserOsDetails = (
+  navigator: Navigator | WorkerNavigator
+): string => {
+  let os = "unknown";
+  const clientStrings = [
+    { s: "Windows 10", r: /(Windows 10.0|Windows NT 10.0)/ },
```
cleve-fauna, line 161: windows 11 is the latest windoes and is not present in this list

package.json
```diff
@@ -20,6 +20,7 @@
     "@tsconfig/node18": "^1.0.1",
     "@types/jest": "^29.4.0",
     "@types/node": "^18.13.0",
+    "@types/serviceworker": "^0.0.67",
```
cleve-fauna, line 23: We don't seem to actually use this package.

cleve-fauna (commented):
Some suggestions.

package.json
ptpaterson: This one is necessary for Typescript to understand the `ServiceWorkerGlobalScope` and `WorkerNavigator` types.

src/util/environment.ts
ptpaterson: good catch! I am looking into this and my understanding is now that it is not possible to distinguish between Windows 10 and 11 through the user agent string. Chromium is freezing user agents strings in favor of newerm methods, which I think means the [User-Agent Client Hints API](https://developer.mozilla.org/en-US/docs/Web/API/User-Agent_Client_Hints_API). Unfortunately though, it's still in an experimental state and at least firefox doesn't support said methods yet.

https://developers.whatismybrowser.com/learn/browser-detection/user-agents/detect-windows-11-user-agent

I suggest we proceed as is, and consider following up with something like:

1. check if the user-agent hints API is available and use that
2. fall back on current impl

package.json
cleve-fauna: where do we use those types? I must have overlooked them. Mind linking me?

cleve-fauna: nm i found one

cleve-fauna (commented):
Two minor suggestions.

cleve-fauna (approved):

PR #159, opened 2023-05-18
long mode

cleve-fauna:
Ticket(s): FE-3538

## Problem
We currently have it set up where

-  BigInts are _always_ encoded as `@long`-tagged values, and
- `@long`-tagged values are _always_ decoded as BigInts.

This makes it cumbersome to use in common cases where BigInts are not strictly necessary.

### Decoding

Customers can configure if '@long`-tagged values are to be decoded as a Number or BigInt at both a client and request level.

If Number is being used, and it's outside of JS’s safe bounds, log a warning using `console.warn`.

### Encoding

If a Number is outside of Fauna’s int boundaries, serialize as a`@long`-tagged value, otherwise serialize as `@int`.

If a BigInt is outside of Fauna’s int boundaries, serialize as a`@long`-tagged value, otherwise serialize as `@int`.

## Testing
Updated tests

----


src/client.ts
```diff
@@ -474,11 +478,13 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
   #validateConfiguration() {
     const config = this.#clientConfiguration;
 
-    const required_options: (keyof ClientConfiguration)[] = [
+    const required_options: (keyof RequiredClientConfig)[] = [
       "client_timeout_buffer_ms",
       "endpoint",
       "format",
       "http2_session_idle_ms",
+      "http2_max_streams",
```
cleve-fauna, line 486: This is already in this array.

```suggestion
```

cleve-fauna (commented):
Three minor suggestions.

PR #162, opened 2023-05-22
Separate QueryOptions from QueryRequest

ptpaterson:
Ticket(s): [FE-3564](https://faunadb.atlassian.net/browse/FE-3564)

## Problem
We do not send query options in the request body, so QueryHeaders is not part of the wire protocol. We've already updated Client.query to only accept the `Query` type, so there is no way for users to provide query options through the current `QueryRequest` type.

## Solution
- `QueryRequest` should not extend `QueryHeaders`
- Rename `QueryHeaders` to `QueryOptions`
- `QueryOptions` should be an optional second argument to query.
- When `QueryOptions` and `ClientConfiguration` make the same setting, `QueryOptions` takes precedence.
- Add an `arguments` field to `QueryOptions` to make up for the fact that we are not passing a `QueryRequest` directly to `Client.query`.

## Result
The wire-protocol module sticks to describing the data can is sent and received.

Types are more accurate and easier to use inside the Client code. It is more clear how the client configuration and per-query options get combined.

## Out of scope
`QueryOptions` should also be passed into `Client.paginate` but that is not done, yet.

## Testing
Updated tests and added a test demonstrating using `arguments`

----

[FE-3564]: https://faunadb.atlassian.net/browse/FE-3564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/integration/query.test.ts
+++ b/__tests__/integration/query.test.ts
@@ -149,6 +149,31 @@ describe("query", () => {
     }
   );
 
+  it("can send arguments directly", async () => {
+    const foo = {
+      double: 4.14,
+      int: 32,
+      string: "foo",
+      null: null,
+      object: { foo: "bar" },
+      array: [1, 2, 3],
+      "@tagged": "tagged",
+    };
+
+    const response = await client.query<typeof foo>(fql`foo`, {
+      arguments: { foo },
+    });
+    const foo2 = response.data;
+
+    expect(foo2.double).toBe(4.14);
+    expect(foo2.int).toBe(32);
+    expect(foo2.string).toBe("foo");
+    expect(foo2.null).toBeNull();
+    expect(foo2.object).toStrictEqual({ foo: "bar" });
+    expect(foo2.array).toStrictEqual([1, 2, 3]);
+    expect(foo2["@tagged"]).toBe("tagged");
+  });
+
   it("throws a QueryCheckError if the query is invalid", async () => {
     expect.assertions(4);
     try {
@@ -422,7 +447,7 @@ describe("query can encode / decode QueryValue correctly", () => {
         Collection.create({ name: ${collectionName}})
       }`);
     // whack in undefined
-    // @ts-ignore
+    // @ts-expect-error Type 'undefined' is not assignable to type 'QueryValue'
     let toughInput: QueryValue = {
       foo: "bar",
       shouldnt_exist: undefined,
@@ -441,7 +466,7 @@ describe("query can encode / decode QueryValue correctly", () => {
   });
 
   it("treats undefined as unprovided passed directly as value", async () => {
-    expect.assertions(2);
+    expect.assertions(3);
     const client = getClient();
     const collectionName = "UndefinedTest";
     await client.query(fql`
@@ -449,7 +474,7 @@ describe("query can encode / decode QueryValue correctly", () => {
         Collection.create({ name: ${collectionName}})
       }`);
     // whack in undefined
-    // @ts-ignore
+    // @ts-expect-error Type 'undefined' is not assignable to type 'QueryValue'
     let undefinedValue: QueryValue = undefined;
     try {
       const docCreated = await client.query(fql`
@@ -462,11 +487,12 @@ describe("query can encode / decode QueryValue correctly", () => {
           }
         })`);
     } catch (e) {
-      if (e instanceof TypeError) {
-        expect(e.name).toBe("TypeError");
+      if (e instanceof ClientError) {
+        expect(e.name).toBe("ClientError");
         expect(e.message).toBe(
-          "Passing undefined as a QueryValue is not supported"
+          "A client level error occurred. Fauna was not called."
         );
+        expect(e.cause).toBeDefined();
       }
     }
   });
--- a/__tests__/integration/query.test.ts
+++ b/__tests__/integration/query.test.ts
@@ -466,7 +466,7 @@ describe("query can encode / decode QueryValue correctly", () => {
   });
 
   it("treats undefined as unprovided passed directly as value", async () => {
-    expect.assertions(3);
+    expect.assertions(2);
     const client = getClient();
     const collectionName = "UndefinedTest";
     await client.query(fql`
@@ -487,12 +487,11 @@ describe("query can encode / decode QueryValue correctly", () => {
           }
         })`);
     } catch (e) {
-      if (e instanceof ClientError) {
-        expect(e.name).toBe("ClientError");
+      if (e instanceof TypeError) {
+        expect(e.name).toBe("TypeError");
         expect(e.message).toBe(
-          "A client level error occurred. Fauna was not called."
+          "Passing undefined as a QueryValue is not supported"
         );
-        expect(e.cause).toBeDefined();
       }
     }
   });
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -1,4 +1,4 @@
-import type { QueryValueObject, ValueFormat } from "./wire-protocol";
+import type { ValueFormat } from "./wire-protocol";
 
 /**
  * Configuration for a client. The options provided are be used as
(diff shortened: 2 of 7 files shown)


src/client-configuration.ts
```diff
@@ -19,6 +20,41 @@ export interface ClientConfiguration {
...
+  /**
+   * When true will keep executing a request even if the page
+   * that fired the request is no longer executing. Only relevant
+   * to underlying clients using the {@link https://fetch.spec.whatwg.org/ | Fetch standard}.
+   * By default set to false.
+   * @remarks
+   * Relevant to clients using the {@link FetchClient} provided,
+   * or any custom HTTP Clients you implement using the Fetch standard.
+   */
+  fetch_keepalive?: boolean;
```
cleve-fauna, line 32: why did we move the lines?

ptpaterson: The fields used for default query options are all under the client-config-only fields

src/client-configuration.ts
```diff
@@ -80,33 +98,74 @@ export interface ClientConfiguration {
...
+}
+
+/**
+ * Options for queries. Each query can be made with different options.
+ */
+export interface QueryOptions {
+  /** Optional arguments. Variables in the query will be initialized to the
+   * value associated with an argument key.
+   */
+  arguments?: QueryValueObject;
```
cleve-fauna, line 122: i don't think this makes sense in `QueryOptions`. This is directly related to text in a `query`. So it should be in `QueryRequest` and nowhere else.

Is this necessary to be here to support using arguments with `fql` formed queries?

If so does passing arguments give us something 

```
fql`${foo}`
``` 

does not?

src/client-configuration.ts
```diff
@@ -1,7 +1,8 @@
-import type { ValueFormat } from "./wire-protocol";
+import type { QueryValueObject, ValueFormat } from "./wire-protocol";
 
 /**
- * Configuration for a client.
+ * Configuration for a client. The query options provided are be used as
+ * default options for each query.
```
cleve-fauna, line 5:
```suggestion
 * Configuration for a client. The options provided are be used as
 * default options for each query.
```

src/client-configuration.ts
```diff
@@ -111,6 +112,64 @@ export interface ClientConfiguration {
   typecheck?: boolean;
 }
 
+/**
+ * Options for queries. Each query can be made with different options.
```
cleve-fauna, line 116:
```suggestion
 * Options for queries. Each query can be made with different options. Settings here
 * take precedence over those in {@link ClientConfiguration}.
```

src/client-configuration.ts
```diff
@@ -111,6 +112,64 @@ export interface ClientConfiguration {
...
+export interface QueryOptions {
+  /** Optional arguments. Variables in the query will be initialized to the
+   * value associated with an argument key.
+   */
+  arguments?: QueryValueObject;
+
+  /**
+   * Determines the encoded format expected for the query `arguments` field, and
+   * the `data` field of a successful response.
+   * Overrides the optional setting for the client.
```
cleve-fauna, line 127:
```suggestion
   * Overrides the optional setting on the {@link ClientConfiguration}.
```

src/client-configuration.ts
```diff
@@ -111,6 +112,64 @@ export interface ClientConfiguration {
...
+  /**
+   * The timeout to use in this query in milliseconds.
+   * Overrides the timeout for the client.
+   */
+  query_timeout_ms?: number;
+
+  /**
+   * A traceparent provided back via logging and telemetry.
+   * Must match format: https://www.w3.org/TR/trace-context/#traceparent-header
+   * Overrides the optional setting for the client.
```
cleve-fauna, line 160:
```suggestion
   * Overrides the optional setting on the {@link ClientConfiguration}.
```

src/client-configuration.ts
```diff
@@ -111,6 +112,64 @@ export interface ClientConfiguration {
...
+   */
+  traceparent?: string;
+
+  /**
+   * Enable or disable typechecking of the query before evaluation. If no value
+   * is provided, the value of `typechecked` in the database configuration will
+   * be used.
+   * Overrides the optional setting for the client.
```
cleve-fauna, line 168:
```suggestion
   * Overrides the optional setting on the {@link ClientConfiguration}.
```

src/client-configuration.ts
```diff
@@ -111,6 +112,64 @@ export interface ClientConfiguration {
   typecheck?: boolean;
 }
 
+/**
+ * Options for queries. Each query can be made with different options.
+ */
+export interface QueryOptions {
```
cleve-fauna, line 118: THis makes more sense in the `wire-protocol` file IMO as they effect the behavior of a query by transmitting things over the wire.

__tests__/integration/query.test.ts
```diff
@@ -462,11 +487,12 @@ describe("query can encode / decode QueryValue correctly", () => {
...
     } catch (e) {
-      if (e instanceof TypeError) {
-        expect(e.name).toBe("TypeError");
+      if (e instanceof ClientError) {
+        expect(e.name).toBe("ClientError");
         expect(e.message).toBe(
-          "Passing undefined as a QueryValue is not supported"
+          "A client level error occurred. Fauna was not called."
         );
+        expect(e.cause).toBeDefined();
```
cleve-fauna, line 495: a `ClientError` implies its our fault - a bug in the driver.

This is a regression if we are now treating bad input as a driver bug.

src/client.ts
```diff
@@ -233,16 +242,16 @@ export class Client {
...
+    options?: QueryOptions
   ): Promise<QuerySuccess<T>> {
     if (this.#isClosed) {
       throw new ClientClosedError(
         "Your client is closed. No further requests can be issued."
       );
     }
 
-    return this.#query(request.toQuery(headers));
+    return this.#query(query, options);
```
cleve-fauna, line 254: i don't think we're ready to make this change yet. 

I suspect this is why we regressed to a `ClientError` from a `TypeError`.

Let's wait on this until we can do this without regressing errors we fire on bad inputs.

src/client.ts
```diff
@@ -30,23 +34,28 @@ import {
...
-  endpoint: URL;
-  format: ValueFormat;
-  http2_session_idle_ms: number;
-  http2_max_streams: number;
-  fetch_keepalive: boolean;
-  secret: string;
-  query_timeout_ms: number;
-}
+type RequiredQueryOptions = QueryOptions &
+  Required<Pick<QueryOptions, "format" | "query_timeout_ms">>;
```
cleve-fauna, line 42: do we need this type? or is typescript smart enough to just infer this when you do the `{ ...clientConfiguration, ...queryOptions }` in `#query`?

src/client-configuration.ts
ptpaterson: There are currently no ways for the user to add arguments, as in `QueryRequest.arguments`. We lost that when we changed the signature of `Client.query` from `(request: QueryRequest | Query, ...)` to (request: Query, ...)`

`QueryRequest` is now just a bit of wire protocol.

cleve-fauna: OK let's leave it here then.

ptpaterson: Didn't mean to be so terse, but I got pulled away. 

I do think that it is an important use case to be able to provide arguments directly. For reference, one thing that comes to mind to me is GraphQL clients accepting "variables". That makes me wonder if we should call name the field `QueryOptions.variables` instead, but I think "arguments" is nice because it's basically a one-to-one what you pass in is sent as `QueryRequest.arguments` (just encoded for tagged format).

I hesitate to try and add it as a third parameter in order to keep `QueryOptions` separate from it.

I am willing to concede we can leave this out now and continue to discussion, I think it could be a clean addition in a later patch. You still okay with this API for arguments?

cleve-fauna: i think its fine how you have it.

src/client.ts
ptpaterson: Yep! By trying to assign a type to `requestConfig` I made it too specific. Like you were getting at, `Client.clientConfiguration` already has these fields set to required, so typescript can figure that out if I just let it infer the type.

__tests__/integration/query.test.ts
ptpaterson: I reverted this back.

src/client-configuration.ts
ptpaterson: Moved it back. Putting it in the "client configuration" file may be a bit of a misnomer. I keep feeling the desire to define the `ClientConfiguration` as an extension of `QueryOptions` since the fields and the types are the same (now they are the same since they are all optional), except... I just added `arguments` field which is now only in `QueryOptions`. Anyways, it still feels like a misfit in `wire-protocol` to me, but it's not something to argue about 🙂

src/client.ts
ptpaterson: Didn't change it all the way back, but did move the `toQuery` up out of `#query`.

src/client-configuration.ts
cleve-fauna: ok we can probably rename the files and put everything in one place in another PR. just thought it woudl be easier if the query is along side its options

src/client-configuration.ts
```diff
@@ -1,7 +1,8 @@
 import type { ValueFormat } from "./wire-protocol";
 
 /**
- * Configuration for a client.
+ * Configuration for a client. The options provided are be used as
```
cleve-fauna, line 4:
```suggestion
 * Configuration for a client. The options provided are used as the
```

PR #165, opened 2023-05-24
Document timeout configuration

ptpaterson:
Ticket(s): FE-3541

## Problem
We need some clarification around timeout configuration options, their relation to each other, related errors, why `Client.close()` should be used, etc.

## Solution
Update the README

Also fixed the missing export for `FaunaError` since we document using that in the README.

## Result
[preview](https://github.com/fauna/fauna-js/blob/af9655bb659f588d424a25d78ef26ca4670e7cf5/README.md)

## Out of scope
Adding every option and public function to the README. We should generate tsdocs for that.

## Testing
n/a

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,7 +284,7 @@ const client = new Client(config);
 
 ### Timeouts
 
-There are a few different timeout settings that can be configured
+There are a few different timeout settings that can be configured; each comes with a default setting. We recommend that most applications simply stick to the defaults.
 
 #### Query Timeout
 The query timeout is the time, in milliseconds, that Fauna will spend executing your query before aborting with a 503 Timeout error. If a query timeout occurs, the driver will throw an instance of `ServiceTimeoutError`.
@@ -295,7 +295,7 @@ The query timeout can be set using the `ClientConfiguration.query_timeout_ms` op
 const client = new Client({ query_timeout_ms: 20_000 });
 ```
 
-The query timeout can also be set to a different value for each query using the `QueryOptions.query_timeout_ms` option. Doing so overrides the client configuration.
+The query timeout can also be set to a different value for each query using the `QueryOptions.query_timeout_ms` option. Doing so overrides the client configuration when performing this query.
 
 ```javascript
 const response = await client.query(myQuery, { query_timeout_ms: 20_000 });
@@ -304,7 +304,7 @@ const response = await client.query(myQuery, { query_timeout_ms: 20_000 });
 #### Client Timeout
 The client timeout is the time, in milliseconds, that the client will wait for a network response before canceling the request. If a client timeout occurs, the driver will throw an instance of `NetworkError`.
 
-The client timeout is always the query timeout plus an additional buffer. This ensures that the client always weights for at least as long Fauna could work on your query and account for network latency. The client timeout buffer is configured by setting the `client_timeout_buffer_ms` option. The default value for the buffer if you do not provide on is 5000 ms (5 seconds), therefore the default client timeout is 10000 ms (10 s) when considering the default query timeout.
+The client timeout is always the query timeout plus an additional buffer. This ensures that the client always waits for at least as long Fauna could work on your query and account for network latency. The client timeout buffer is configured by setting the `client_timeout_buffer_ms` option. The default value for the buffer if you do not provide on is 5000 ms (5 seconds), therefore the default client timeout is 10000 ms (10 s) when considering the default query timeout.
 
 ```javascript
 const client = new Client({ client_timeout_buffer_ms: 6000 });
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -213,23 +232,27 @@ import { fql, Client, type QueryOptions } from "fauna";
...
   query_tags: { name: "readme query" },
+  query_timeout_ms: 60_000,
   traceparent: "00-750efa5fb6a131eb2cf4db39f28366cb-000000000000000b-00",
   typecheck: true,
 };
 
-const result = await client.query(fql`"Hello, Fauna!"`, options);
+const response = await client.query(fql`"Hello, ${name}!"`, options);
+
+client.close()
```
ptpaterson, line 248:
```suggestion
client.close();
```

README.md
```diff
@@ -238,22 +261,73 @@ import { Client, endpoints, type ClientConfiguration } from "fauna";
...
+The query timeout can also be set to a different value for each query using the `QueryOptions.query_timeout_ms` option. Doing so overrides the client configuration.
+
+```javascript
+const response = await client.query(myQuery, { query_timeout_ms: 20_000 });
+```
+
+#### Client Timeout
+The client timeout is the time, in milliseconds, that the client will wait for a network response before canceling the request. If a client timeout occurs, the driver will throw an instance of `NetworkError`.
+
+The client timeout is always the query timeout plus an additional buffer. This ensures that the client always weights for at least as long Fauna could work on your query and account for network latency. The client timeout buffer is configured by setting the `client_timeout_buffer_ms` option. The default value for the buffer if you do not provide on is 5000 ms (5 seconds), therefore the default client timeout is 10000 ms (10 s) when considering the default query timeout.
```
cleve-fauna, line 307:
```suggestion
The client timeout is always the query timeout plus an additional buffer. This ensures that the client always waits for at least as long Fauna could work on your query and account for network latency. The client timeout buffer is configured by setting the `client_timeout_buffer_ms` option. The default value for the buffer if you do not provide on is 5000 ms (5 seconds), therefore the default client timeout is 10000 ms (10 s) when considering the default query timeout.
```

README.md
```diff
@@ -238,22 +261,73 @@ import { Client, endpoints, type ClientConfiguration } from "fauna";
...
+#### Query Timeout
+The query timeout is the time, in milliseconds, that Fauna will spend executing your query before aborting with a 503 Timeout error. If a query timeout occurs, the driver will throw an instance of `ServiceTimeoutError`.
+
+The query timeout can be set using the `ClientConfiguration.query_timeout_ms` option. The default value if you do not provide one is 5000 ms (5 seconds).
+
+```javascript
+const client = new Client({ query_timeout_ms: 20_000 });
+```
+
+The query timeout can also be set to a different value for each query using the `QueryOptions.query_timeout_ms` option. Doing so overrides the client configuration.
```
cleve-fauna, line 298:
```suggestion
The query timeout can also be set to a different value for each query using the `QueryOptions.query_timeout_ms` option. Doing so overrides the client configuration when performing this query.
```

README.md
```diff
@@ -238,22 +261,73 @@ import { Client, endpoints, type ClientConfiguration } from "fauna";
...
+
+```javascript
+const client = new Client({ http2_session_idle_ms: 6000 });
+```
+
+> **Note**
+> Your application process may continue executing after all requests are completed for the duration of the session idle timeout. To prevent this, it is recommended to call `Client.close()` once all requests are complete. It is not recommended to set `http2_session_idle_ms` to small values.
+
+> **Warning**
+> Setting `http2_session_idle_ms` to small values can lead to a race condition where requests cannot be completed and you received a `ERR_HTTP2_GOAWAY_SESSION` error.
```
cleve-fauna, line 328:
```suggestion
> Setting `http2_session_idle_ms` to small values can lead to a race condition where requests cannot be transmitted before the session is closed, yielding `ERR_HTTP2_GOAWAY_SESSION` errors.
```

README.md
```diff
@@ -238,22 +261,73 @@ import { Client, endpoints, type ClientConfiguration } from "fauna";
...
   traceparent: "00-750efa5fb6a131eb2cf4db39f28366cb-000000000000000b-00",
   typecheck: true,
 };
 
 const client = new Client(config);
 ```
 
+### Timeouts
+
+There are a few different timeout settings that can be configured
```
cleve-fauna, line 287:
```suggestion
There are a few different timeout settings that can be configured; each comes with a default setting. We recommend that most applications simply stick to the defaults.
```

cleve-fauna (approved):

README.md
ptpaterson: ope!!

PR #169, opened 2023-06-05
Handle missing process.env correctly

ptpaterson:
Ticket(s): FE-3607

## Problem
We use `process.env` to fallback on env vars for client secret, but we are blowing up in environments where `process.env` is undefined.

A customer using Vue & Nuxt encountered an issue where creating a `Client` instance breaks Vue reactivity. The issue is triggered because `process` is defined to be something like this:

```javascript
{ dev: true, test: false, server: false, client: true }
```

## Solution
Check for `process.env` is what we actually want

## Result
We can use the driver with Nuxt.js

## Out of scope
Adding Nuxt.js platform testing.

## Testing
Local testing with a minimal Vue.js app. Create the simplest app:

```shell
npx nuxi@latest init my-app
```

Update the `app.vue` file:

```vue
<template>
  <div>
    <input type="text" v-model="title" />

    {{ title }}
    {{ data }}
  </div>
</template>

<script setup>

const title = ref('');
const data = ref('')

import { Client, fql } from "fauna";

// creating the client will break the app before this fix
const client = new Client({
  secret: YOUR_FAUNA_KEY,
});

client.query(fql`"hello world"`).then(res => data.value = res.data)
</script>
```

start the app and see the request complete and reactivity when you type into the input field.

```shell
npm run dev
```

if linking installing a local version of the driver, Nuxt doesn't like symlinks, so add this to nuxt.config.ts

```javascript
  vite: {
    server: {
      fs: {
        allow: ['/Users/psquared/Documents/fauna/fauna-js'],
      },
    },
  },
  imports: {
    transform: {
      exclude: [/\bfauna\b/]
    }
  },
```

----


src/client.ts
```diff
@@ -296,10 +296,11 @@ export class Client {
...
 
   #getSecret(partialClientConfig?: ClientConfiguration): string {
-    let fallback = undefined;
-    if (typeof process === "object") {
-      fallback = process.env["FAUNA_SECRET"];
-    }
+    const fallback =
+      process?.env !== null && typeof process.env === "object"
+        ? process.env["FAUNA_SECRET"]
+        : undefined;
```
cleve-fauna, line 302: can we just do this?

```suggestion
    const fallback = process?.env?.FAUNA_SECRET;
```

src/util/environment.ts
```diff
@@ -240,7 +240,7 @@ const crossGlobal =
  */
 const getNodeRuntimeEnv = (): string => {
   // return early if process variables are not available
-  if (!(typeof process !== "undefined" && process && process.env)) {
+  if (!(process?.env !== null && typeof process.env === "object")) {
```
cleve-fauna, line 243:
```suggestion
  if (typeof process?.env !== "object") {
```

cleve-fauna (commented):
some simplifications

cleve-fauna (commented):
some simplifications

src/util/environment.ts
```diff
@@ -240,7 +240,7 @@ const crossGlobal =
  */
 const getNodeRuntimeEnv = (): string => {
   // return early if process variables are not available
-  if (!(typeof process !== "undefined" && process && process.env)) {
+  if (typeof process?.env !== "object") {
```
cleve-fauna, line 243:
```suggestion
  if (typeof process?.env !== "object" || process.env === null) {
```

ptpaterson:
yeeeeea that was much simpler.

src/util/environment.ts
cleve-fauna: null's type is object

cleve-fauna (approved):
Ship if you've integ tested using the process you described in the description.

PR #174, opened 2023-07-10
add schema version to drivers

fauna-chase:
Leaving this as draft for now as the change to add the schema version hasn't made it out to prod yet.

ENG-5237
This builds on a change to core that started including the schema version in the query API response: https://github.com/fauna/core/blob/master/ext/api/src/main/resources/openapi.yml#L393 This schema version can be used by the dashboard to determine when it needs to refresh the schema.  This includes both displaying schema to users as well as refreshing the environment used by the language library to provide intellisense.

### DIFF
--- a/__tests__/integration/schema-version.test.ts
+++ b/__tests__/integration/schema-version.test.ts
@@ -5,13 +5,13 @@ describe("schema version is returned by the client", () => {
   const client = getClient();
   it("returns the schema version", async () => {
     const resTs = await client.query(fql`
-    Collection.create({ name: "TestColl" })
+      Collection.create({ name: "TestColl" })
     `);
 
     const expectedSchemaVersion = resTs.txn_ts;
 
     const res = await client.query(fql`
-    Customers.all()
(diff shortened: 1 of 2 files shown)


__tests__/integration/schema-version.test.ts
```diff
@@ -0,0 +1,19 @@
+import { fql } from "../../src";
+import { getClient } from "../client";
+
+describe("schema version is returned by the client", () => {
+  const client = getClient();
+  it("returns the schema version", async () => {
+    const resTs = await client.query(fql`
+    Collection.create({ name: "TestColl" })
+    `);
```
macmv, line 9: looks like the rest of their tests indent these:
```suggestion
    const resTs = await client.query(fql`
      Collection.create({ name: "TestColl" })
    `);
```

macmv (approved):
Aside from formatting LGTM

ptpaterson (approved):

macmv (approved):

PR #178, opened 2023-07-12
fix: Explicitly check for browser env

### LINKED ISSUE #173: Support Vercel Edge environment
Hey folks, I'm developing an app using NextJS and I have a function that runs on the edge, and when I try to use the fauna client in this function, I get the following error:
```
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

Import trace for requested module:
./node_modules/fauna/dist/browser/index.js
Error [ReferenceError]: navigator is not defined
```
I guess it is because it tries to access some browser feature that is not available 🤔 

To reproduce this you create a NextJS project and inside of the app folder, create a `/api/route.ts` with something like:
```ts
export const runtime = "edge";

export function GET() {
    const fauna = new Client({ ... })
    fauna.query(fql`[]`)
}
```
Sorry for not having too much more info to help with debug 😞

ptpaterson:
Ticket(s): [FE-3729](https://faunadb.atlassian.net/browse/FE-3729)

Fixes #173 

Leaving this as a draft while we get some testing done.

## Problem
We currently assume that if you are not running in Node or Service Worker, you are in a browser. That blows up in environments like Vercel's Edge Runtime and Fastly's Compute@Edge.

## Solution
Explicitly check for browser env before getting details from `navigator` object. If none of our env checks are true, then report unknown.

## Testing
Running locally with some edge projects.

----

[FE-3729]: https://faunadb.atlassian.net/browse/FE-3729?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/util/environment.ts
+++ b/src/util/environment.ts
@@ -19,14 +19,37 @@ export const getDriverEnv = (): string => {
     runtime: "unknown",
   };
 
-  let isServiceWorker = false;
-  try {
-    isServiceWorker = global instanceof ServiceWorkerGlobalScope;
-  } catch (_) {
-    // okay if this fails
-  }
+  /**
+   * Determine if we're executing in a Node environment
+   */
+  const isNode =
+    typeof window === "undefined" &&
+    typeof process !== "undefined" &&
+    process.versions != null &&
+    process.versions.node != null;
 
-  if (isNode()) {
+  /**
+   * Determine if we're executing in a Node environment
+   */
+  const isBrowser =
+    typeof window !== "undefined" && typeof window.document !== "undefined";
+
+  /**
+   * Determine if we're executing in a Service Worker environment
+   */
+  const isServiceWorker =
+    typeof self === "object" &&
+    self.constructor &&
+    self.constructor.name === "DedicatedWorkerGlobalScope";
+
+  /**
+   * Determine if we're executing in Vercel's Edge Runtime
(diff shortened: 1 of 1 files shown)


cleve-fauna:
seems OK without hardcore testing.

src/util/environment.ts
```diff
@@ -19,25 +19,51 @@ export const getDriverEnv = (): string => {
     runtime: "unknown",
   };
 
-  let isServiceWorker = false;
-  try {
-    isServiceWorker = global instanceof ServiceWorkerGlobalScope;
-  } catch (_) {
-    // okay if this fails
-  }
```
cleve-fauna, line 27: are we losing some safety?

cleve-fauna (commented):
Wrapping this whole process in some kind of try catch so we don't blow up on writing this data to headers would be nice if we don't have it already.

src/util/environment.ts
ptpaterson: No, we should be fine. I found recommendations for checking for Service Worker env that don't rely on trying something that can throw. This try/catch block was only around using `instanceof`. `typeof` doesn't throw an error if the variable isn't declared.

cleve-fauna (dismissed):

cleve-fauna (commented):
Discussed on slack - this telemetry could still cause the driver to be unusuable in idiosyncratic envs. Plan is to wrap in try catch to prevent that.

cleve-fauna (approved):

PR #182, opened 2023-07-18
fix schema version test

fauna-chase:
My original test was working locally but I didn't' correctly account for resource management when running in the pipeline.  This should resolve that.

### DIFF
--- a/__tests__/integration/schema-version.test.ts
+++ b/__tests__/integration/schema-version.test.ts
@@ -5,13 +5,17 @@ describe("schema version is returned by the client", () => {
   const client = getClient();
   it("returns the schema version", async () => {
     const resTs = await client.query(fql`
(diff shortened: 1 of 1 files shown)


__tests__/integration/schema-version.test.ts
```diff
@@ -5,13 +5,17 @@ describe("schema version is returned by the client", () => {
   const client = getClient();
   it("returns the schema version", async () => {
     const resTs = await client.query(fql`
-      Collection.create({ name: "TestColl" })
+      if (Collection.byName("TestColl") == null) {
+        Collection.create({ name: "TestColl" })
+      } else {
+        TestColl.definition.update({})
```
ptpaterson, line 11: Would this work? To avoid the write?

```suggestion
        TestColl.definition
```

ptpaterson (approved):

PR #183, opened 2023-07-19
Add test helpers

macmv:
Adds a test helper we have in the database (`newDB`), which creates a fresh DB for each run.

It then uses this helper to fix the schema version test.

### DIFF
--- a/__tests__/integration/schema-version.test.ts
+++ b/__tests__/integration/schema-version.test.ts
@@ -1,15 +1,18 @@
 import { fql } from "../../src";
-import { getClient } from "../client";
+import { newDB } from "../client";
+import { Client } from "../../src";
 
 describe("schema version is returned by the client", () => {
-  const client = getClient();
+  let client: Client;
+
+  // make a fresh db each run
+  beforeAll(async () => {
+    client = await newDB("SchemaTest");
+  });
+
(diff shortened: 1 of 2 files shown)


__tests__/integration/schema-version.test.ts
```diff
@@ -1,15 +1,17 @@
...
+    client = await newDB("SchemaTest");
+  });
+
   it("returns the schema version", async () => {
     const resTs = await client.query(fql`
-      if (Collection.byName("TestColl") == null) {
-        Collection.create({ name: "TestColl" })
-      } else {
-        TestColl.definition.update({})
-      }
```
ptpaterson, line 12: Did this not work because `.update({})` is effectively a no-op? What if we just incremented some meta-data? Would changing the `data` not change the schema version.

macmv: I'm not sure if it would change the schema version. Its easier to have a clean DB each time the test is ran, so I just went ahead and wrote the helper.

If you'd prefer to mess around with `update` I can change the PR, I just thought that making a fresh DB would lead to more reliable and simpler tests.

ptpaterson (approved):

fauna-chase (approved):

Issue #199, opened 2023-08-26
error TS2380


lenanex:
Hello,

I've installed the library in a Ionic Project and when I'm importing fauna as per the documentation (import { Client, fql, FaunaError } from "fauna") I'm getting a typescript compilation error:

 Error: node_modules/fauna/dist/client.d.ts:30:9 - error TS2380: The return type of a 'get' accessor must be assignable to its 'set' accessor type
[ng]  Type 'undefined' is not assignable to type 'number'.
[ng] 
[ng] 30     get lastTxnTs(): number | undefined;

Thanks a lot for your help :)

cleve-fauna:
Fix is in fauna-js 1.0.2 https://www.npmjs.com/package/fauna

PR #200, opened 2023-08-29
Accept undefined in setLastTxnTs input but treat it as a no-op. This resolves typescript compilation errors while maintaining current behavior

cleve-fauna:
Ticket(s): #199 
## Problem
We are producing invalid `client.d.ts` files as the getter outputs different types for the setter for the field `lastTxnTs`.
## Solution
Accept the same types for input to the `lastTxnTs` setter as are output from the getter. Maintain current behavior of the setter by ignoring undefined inputs (so users cannot break read their own write by passing in undefined here).
## Result
No more typescript errors.
## Testing
Running regression tests via GitHub.

----


__tests__/integration/client-last-txn-tracking.test.ts
```diff
@@ -98,5 +98,29 @@ describe("last_txn_ts tracking in client", () => {
     `);
     expect(resultThree.txn_ts).not.toBeUndefined();
     expect(resultThree.txn_ts).not.toEqual(resultTwo.txn_ts);
+    myClient.close();
+  });
+
+  it("Ingores overrides  of the lastTxnTs that are less than the current value or undefined", async () => {
```
cleve-fauna, line 104:
```suggestion
  it("Ignores overrides  of the lastTxnTs that are less than the current value or undefined", async () => {
```

henryfauna (approved):

PR #204, opened 2023-08-29
Pin fauna version in pipeline

cleve-fauna:
## Problem
The pipeline is stuck because we cannot run integration tests successfully due to compatibility issues in the docker images we use inside our CD pipeline.
## Solution
A temporary work around for this is to pin the fauna db image we use in the CD pipeline to 4.28
## Result
We can run the pipeline.

## Testing
Made this same work around on the pipeline for a private repo and it works as expected.

----

### DIFF
--- a/concourse/scripts/docker-compose-fauna.yml
+++ b/concourse/scripts/docker-compose-fauna.yml
@@ -2,7 +2,7 @@ version: "3.3"
 
 services:
   faunadb:
-    image: fauna/faunadb
+    image: fauna/faunadb:4.28
     container_name: faunadb
     ports:
       - "8443:8443"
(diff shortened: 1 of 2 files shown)


concourse/scripts/docker-compose-fauna.yml
```diff
@@ -2,7 +2,7 @@ version: "3.3"
 
 services:
   faunadb:
-    image: fauna/faunadb
+    image: fauna/faunadb:4.28
```
cleve-fauna, line 5:
```suggestion
    image: fauna/faunadb:4.28.0
```

sprsquish (approved):

Issue #205, opened 2023-08-30
FAUNA_ENDPOINT Environment Variable Is Not Honored


andyvanosdale:
In the README, it states that the `FAUNA_ENDPOINT` environment variable can be used to override the default endpoint client configuration setting. However, this is not getting honored when set.

From what I see in the code, this just has not been implemented yet. There is an endpoint configuration that does have locahost settings, so this can be worked around with:

```
const client = process.env.FAUNA_ENDPOINT : 
  new Client({
    endpoint: endpoints.local
  }) ?
  new Client()
```

Repro Steps:
index.ts
```ts
import { Client } from 'fauna'
const client = new Client()
console.log(client.clientConfiguration)
```
Terminal
```
export FAUNA_ENDPOINT=http://localhost:8443
# transpile index.ts to javascript
node index.js
```

The output will show the default endpoint configuration (https://db.fauna.com/) instead of the environment setting (http://localhost:8443).

CLBray:
Thanks for this report, @andyvanosdale! We'll get this fixed up shortly.

PR #206, opened 2023-08-30
endpoint falls back to FAUNA_ENDPOINT env variable

ptpaterson:
Ticket(s): FE-4347

fixes: #205

## Problem
In the README, it states that the FAUNA_ENDPOINT environment variable can be used to override the default endpoint client configuration setting. However, this is not getting honored when set.

## Solution
Fall back on the env variable if endpoint is not explicitly set. THEN fall back on the default (https://db.fauna.com).

If the endpoint is explicitly set to `undefined`, then we still throw a TypeError, rather than fall back on any values.

## Result
env var is honored.

## Testing
Added a test.

----

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -55,10 +55,9 @@ type RequiredClientConfig = ClientConfiguration &
 
 const DEFAULT_CLIENT_CONFIG: Omit<
   ClientConfiguration & RequiredClientConfig,
-  "secret"
+  "secret" | "endpoint"
 > = {
   client_timeout_buffer_ms: 5000,
-  endpoint: endpoints.default,
   format: "tagged",
   http2_session_idle_ms: 5000,
   http2_max_streams: 100,
@@ -106,6 +105,7 @@ export class Client {
       ...DEFAULT_CLIENT_CONFIG,
       ...clientConfiguration,
       secret: this.#getSecret(clientConfiguration),
+      endpoint: this.#getEndpoint(clientConfiguration),
     };
 
     this.#validateConfiguration();
@@ -295,7 +295,7 @@ export class Client {
   }
 
   #getSecret(partialClientConfig?: ClientConfiguration): string {
-    let fallback = undefined;
+    let env_secret = undefined;
     if (
       typeof process !== "undefined" &&
       process &&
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -317,6 +317,35 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+    // TypeError, rather than override with the default endpoint.
+    if (
+      partialClientConfig &&
+      "endpoint" in partialClientConfig &&
+      partialClientConfig.endpoint === undefined
+    ) {
+      throw new TypeError(
+        `ClientConfiguration option endpoint must be defined.`
+      );
+    }
```
pnwpedro, line 331: I actually think we shouldn't care if someone sets it to null or undefined. If they do, let it coalesce to the env var or the default value.

If you do think this is important, why not disallow null as well?

pnwpedro (changes requested):

src/client.ts
ptpaterson: We've been here before 😄 I ported over an `applyDefaults` function from the v4 driver that treated explicit undefined values the same as not-provided, thus applying defaults on top of them. 

https://github.com/fauna/fauna-js/pull/145#discussion_r1187792805

And in response: 

> Treating undefined special makes the interface more complex

https://github.com/fauna/fauna-js/pull/145#discussion_r1188797204

What we get by not accepting explicitly set undefined values is we catch what are most likely (if, admittedly, not always) mistakes/bugs in user code.

Regarding `null`: there are plenty of other bad inputs one could provide, including `null`, but we took the path of focusing on enforcing through TS types. Our types are actually currently enforcing that provided values must be defined (we specify `field?: Type` and not `field?: Type | undefined`), as well as not `null`. But it's a really common JS bug that assumes a variable is defined but isn't. It it more rare to accidentally assign a variable to null when it should have a value.

All that said, I do not mean to argue very strongly for keeping the errors for explicit undefined -- I'm really just trying to lay out the history for the record. I think this is a design choice where neither option is more "correct".

It would technically be a breaking change to start overriding explicit undefined values. The types disallow explicit undefined, so what we would be doing is changing an explicit error to one we handle internally for the user by applying our own defaults.

pnwpedro: Okay, if we've been here we've been here :) I'm just in favor of less code, but I'll ship this.

If it comes up again, hopefully I'll remember.

pnwpedro (approved):

PR #207, opened 2023-08-31
Allow overriding the client's secret

macmv:
Ticket(s): FE-4356

This allows overriding the client's secret on a `query` call.

This might screw up `last_txn_ts` if you switch region groups. That should quickly resolve itself, as the last_txn_ts will catch up quickly.

### DIFF
--- a/__tests__/unit/client.test.ts
+++ b/__tests__/unit/client.test.ts
@@ -13,8 +13,8 @@ import {
 
 describe("Client", () => {
   it("Allows setting a secret in query", async () => {
-    expect.assertions(1)
-  
+    expect.assertions(1);
+
     const client = getClient();
     try {
       await client.query(fql`Role.create({ name: "hi" })`, {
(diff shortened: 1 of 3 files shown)


ptpaterson (changes requested):
Can you add a test for this? Otherwise, LGTM.

__tests__/unit/client.test.ts
```diff
@@ -1,11 +1,35 @@
...
+} from "../../src";
 import {
   getClient,
   getDefaultHTTPClientOptions,
   getDefaultSecretAndEndpoint,
 } from "../client";
 
 describe("Client", () => {
+  it("Allows setting a secret in query", async () => {
+    const client = getClient();
```
ptpaterson, line 16: Make sure the test fails if a different error occurs.

```suggestion
  it("Allows setting a secret in query", async () => {
    expect.assertions(1)
  
    const client = getClient();
```

ptpaterson (approved):

PR #208, opened 2023-08-31
Expose the secret in `client.clientConfiguration`

macmv:
Ticket(s): FE-###

I'm not sure why we hid this in the first place. The code to construct the client needs to know the secret, and this just lets you avoid passing the secret around alongside the client.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -141,12 +141,11 @@ export class Client {
   }
 
   /**
-   * Return the {@link ClientConfiguration} of this client, save for the secret.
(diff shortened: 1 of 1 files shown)


ptpaterson:
The configuration is private to prevent accidentally logging the secret (following git blame https://github.com/fauna/fauna-js/pull/55). But I agree that the getter method should be able to return it. Getters don't get printed when logged or serialized with JSON.stringify.

ptpaterson (approved):

PR #211, opened 2023-09-05
Add retries on throttling errors

pnwpedro:
Ticket(s): FE-2988

## Problem
We need to retry throttling errors within the client.

## Solution
* Expose max_attempts and max_backoff in client configuration
* Backoff/retry on throttling error up to max_attempts

## Testing
* Added unit test

----

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -50,6 +50,8 @@ type RequiredClientConfig = ClientConfiguration &
       | "format"
       | "long_type"
       | "query_timeout_ms"
+      | "max_attempts"
+      | "max_backoff"
     >
   >;
 
@@ -64,6 +66,8 @@ const DEFAULT_CLIENT_CONFIG: Omit<
   long_type: "number",
   fetch_keepalive: false,
   query_timeout_ms: 5000,
+  max_attempts: 3,
+  max_backoff: 20,
 };
 
 /**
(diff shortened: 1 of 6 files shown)


src/client.ts
```diff
@@ -254,7 +261,32 @@ export class Client {
...
+    const maxBackoff =
+      this.clientConfiguration.max_backoff ?? DEFAULT_CLIENT_CONFIG.max_backoff;
+    const maxAttempts =
+      this.clientConfiguration.max_attempts ??
+      DEFAULT_CLIENT_CONFIG.max_attempts;
+    const backoffMs =
+      Math.min(Math.random() * 2 ** attempt, maxBackoff) * 1_000;
+
+    const wait = (ms: number) => new Promise((r) => setTimeout(r, ms));
+    attempt += 1;
```
ptpaterson, line 281: Is a random number preferred here? That could mean any retry could be near 0s.

We could ramp up the exponential backoff as a function ramping up to equal `max_backoff` at `max_attempts`

```suggestion
    // backoff formula:  t = e ^ (r/c)
    // where t is the backoff time, r is retry attempt, c is a constant
    // t_max = e ^ (r_max/c)  =>  c = r_max/ln(t_max)
    
    attempt += 1;
    const c = (maxAttempts - 1) / Math.log(maxBackoff)
    const backoffMs = Math.exp(attempt/c) * 1_000;

    const wait = (ms: number) => new Promise((r) => setTimeout(r, ms));
```

pnwpedro: Yeah, this is the impl we've chosen in all drivers.

ptpaterson: okie dokie then.

ptpaterson (approved):

PR #213, opened 2023-09-07
Allow setting arbitrary headers

macmv:
Ticket(s): FE-4378

Allows setting arbitrary headers in the initial config and in `query`. These headers will overwrite any headers the driver sets up.


macmv:
see convo in slack, don't want to support this right now

PR #214, opened 2023-09-18
Support users embedding fql snippets in objects and arrays.

ptpaterson:
Ticket(s): FE-3501

## Problem
`Query` instances (i.e. results from the `fql` function) cannot be encoded when embedded in objects or arrays.

## Solution
The database now accepts a query interpolation fragments for `array` and `object`, where the existing `fql` fragment can now be nested under the `array` and `object`. Add new `encodeInterpolation` method to encode with the new fragments.

Query `arguments` do not use interpolation, so they still need the `@object` tag. Encoding arguments still uses the existing `encode` method.

Remove encoding of Queries from the `Query` class itself. Encoding of arguments cannot handle Queries.

## Result
Queries can be encoded in QueryInterpolations.

Queries can be embedded within JS arrays and objects.

## Out of scope
n/a

## Testing
Updated existing tests and added new ones for embedding queries in arrays and objects.

----


__tests__/unit/query-builder.test.ts
```diff
@@ -132,38 +150,7 @@ describe("fql method producing Querys", () => {
         },
       ],
     });
-    expect(queryRequest.arguments).toStrictEqual({});
-  });
-
-  it("adds headers if passed in", () => {
```
cleve-fauna, line 138: why was this test deleted?

src/query-builder.ts
```diff
@@ -36,8 +36,8 @@ export function fql(
  * function
  */
 export class Query {
-  readonly #queryFragments: ReadonlyArray<string>;
-  readonly #queryArgs: (QueryValue | Query)[];
+  readonly queryFragments: ReadonlyArray<string>;
+  readonly queryArgs: (QueryValue | Query)[];
```
cleve-fauna, line 40: Seems pretty leaky to expose this to users. How about we pass these as arguments to TaggedTypeFormat.encodeInterpolation so these can stay private.

cleve-fauna, line 40: By moving encoding to strictly the TaggedTyped file we've now made these previously private fields public.

That feels like a regression as we don't want users to take any dependencies on these.

IMO it would be better to keep these private even it means we don't have as clean a break between files as we'd like.

Are there reasons we don't want to do some encoding here other than modular purity?

__tests__/unit/tagged-format.test.ts
```diff
@@ -142,143 +143,170 @@ describe.each`
...
-    // Set types
-    // TODO: uncomment to add test once core accepts `@set` tagged values
-    // expect(backToObj.page).toStrictEqual({ "@set": { data: ["a", "b"] } });
-    // TODO: uncomment to add test once core accepts `@set` tagged values
-    // expect(backToObj.page_string).toStrictEqual({ "@set": "abc123" });
   });
 
   it("handles conflicts", () => {
-    const result = TaggedTypeFormat.encode({
+    const result = TaggedTypeFormat.encodeInterpolation({
```
macmv, line 255: is the `@object` behavior still tested? From what I can tell, it looks like all these tests just got refactored, and now the `@object` behavior isn't tested.

src/tagged-type.ts
```diff
@@ -253,9 +284,109 @@ const encode = (input: QueryValue): QueryValue => {
...
+  }
+  switch (typeof input) {
+    case "bigint":
+      return encodeValueInterpolation(encodeMap["bigint"](input));
+    case "string":
+      return encodeValueInterpolation(encodeMap["string"](input));
+    case "number":
+      return encodeValueInterpolation(encodeMap["number"](input));
+    case "boolean":
+      return encodeValueInterpolation(input);
```
macmv, line 310: can you do a `default: return encode(input)` or something?

src/tagged-type.ts
```diff
@@ -253,9 +284,109 @@ const encode = (input: QueryValue): QueryValue => {
...
+      } else if (input instanceof NamedDocumentReference) {
+        return encodeValueInterpolation(
+          encodeMap["namedDocumentReference"](input)
+        );
+      } else if (input instanceof NullDocument) {
+        return encodeInterpolation(input.ref);
+      } else if (input instanceof Page) {
+        return encodeValueInterpolation(encodeMap["set"](input));
+      } else if (input instanceof EmbeddedSet) {
+        return encodeValueInterpolation(encodeMap["set"](input));
```
macmv, line 339: same for all of this. it seems like theres a lot duplicated between this and `encode`.

__tests__/integration/query.test.ts
```diff
@@ -166,6 +166,7 @@ describe("query", () => {
 
     const response = await client.query<typeof foo>(fql`foo`, {
       arguments: { foo },
+      format: "tagged",
```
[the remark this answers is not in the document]
ptpaterson: Quite possibly not necessary. I'll check and remove if not. I was trying different things while tests were failing.

__tests__/unit/tagged-format.test.ts
ptpaterson: The diff is pretty gnarly now, but I put the original tests back together and re-titled the new ones for `encodeInterpolation`

src/tagged-type.ts
ptpaterson: There is a small, finite list of cases: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#description

The only cases left were "symbol" and "function", which I just added for completeness. Thanks for catching!

__tests__/integration/template-format.test.ts
```diff
@@ -109,6 +109,36 @@ describe("query using template format", () => {
     expect(response.data).toBe(true);
   });
 
+  it("succeeds with expressions nested within objects", async () => {
```
cleve-fauna, line 112: i would like to see at a test for mulitple layers of nesting. Like:

const arg = {
  a: fql`1 + ${fql`2`}`
}
queryBuilder = fql`${arg}`

cleve-fauna: i've added such a test

cleve-fauna (dismissed):
Looks good to me but want @macmv sign off before merge since i had a hand in the code here.

cleve-fauna (dismissed):

src/query-builder.ts
```diff
@@ -36,8 +37,8 @@ export function fql(
  * function
  */
 export class Query {
-  readonly #queryFragments: ReadonlyArray<string>;
-  readonly #queryArgs: (QueryValue | Query)[];
+  #queryFragments: ReadonlyArray<string>;
+  #queryArgs: (QueryValue | Query)[];
```
ptpaterson, line 41: These should still stay readonly.

```suggestion
  readonly #queryFragments: ReadonlyArray<string>;
  readonly #queryArgs: (QueryValue | Query)[];
```

src/tagged-type.ts
```diff
@@ -253,9 +281,79 @@ const encode = (input: QueryValue): QueryValue => {
...
+};
+
+const encodeArrayInterpolation = (input: Array<QueryValue>): ArrayFragment => {
+  const encodedItems = input.map(encodeInterpolation);
+  return { array: encodedItems };
+};
+
+const encodeQueryInterpolation = (value: Query): FQLFragment => {
+  return value.toQuery().query;
+};
```
ptpaterson, line 355:
```suggestion
const encodeQueryInterpolation = (value: Query): FQLFragment =>
  value.toQuery().query;
```

macmv (approved):

PR #232, opened 2024-02-08
Initial Streaming Implementation

ptpaterson:
Ticket(s): FE-4969

This is an initial go at implementing streaming events for the v10 API. Anything here is up for review and comment.

## Problem
Implement the v10 Fauna streaming API

We need to support browsers and Node. We would also like to support other runtimes if possible.

## Solution
Provide a `StreamClient` class which provides the primary user interface to establish a stream and handle events.

Provide a lower-level `HTTPStreamClient` interface that can be implemented for different environments.

The `StreamClient` contains an instance of an `HTTPStreamClient`, which mirrors the existing lower-level `HTTPClient` interface that enables different backend implementations for the `Client` class.

The `HTTPStreamClient` is responsible for initiating a request and returning an object that implements the `StreamAdapter` interface. The `StreamAdapter` contains an asyncronous generator function that yields Fauna streaming events as raw strings for the `StreamClient` to use. The `StreamAdapter` also provides a `close` method to provide a mechanism to close the stream.

The `StreamClient` is responsible for parsing the raw-string events into valid types. Streaming events always return data in the "tagged" format. The `StreamClient` must be configured with how to handle decoding of `Long` types.

An instance of `StreamClient` is created on each call to `Client.stream`. Each instance tracks it's own handlers for events. The underlying HTTP connections should be reused when possible.

## Result
Streams can be used in the browser and Node.js.

This does not work in public Fauna environments or the docker image yet.

### Usage

Two ways to use the stream are enabled: 
1. Async Iterator
2. Callbacks

_Async Iterator example_
```javascript
import { Client, fql } from "fauna"
const client = new Client({ secret: FAUNA_SECRET })

const response = await client.query(fql`MyCollection.all().toStream()`);
const streamToken = response.data;

const stream = client.stream(streamToken)

try {
    switch (event.type) {
      case "update":
      case "add":
      case "remove":
        console.log("Stream update:", event);
        // ...
        break;
    }
  }
} catch (error) {
  // An error will be handled here if Fauna returns a terminal, "error" event, or
  // if Fauna returns a non-200 response when trying to connect, or
  // if the max number of retries on network errors is reached.

  // ... handle fatal error
}
```

_Callbacks example_
```javascript
import { Client, fql } from "fauna"
const client = new Client({ secret: FAUNA_SECRET })

const response = await client.query(fql`MyCollection.all().toStream()`);
const streamToken = response.data;

const stream = client.stream(streamToken)
  
stream.start(
  function onEvent(event) {
    switch (event.type) {
      case "update":
      case "add":
      case "remove":
        console.log("Stream update:", event);
        // ...
        break;
    }
  },
  function onFatalError(error) {
    // An error will be handled here if Fauna returns a terminal, "error" event, or
    // if Fauna returns a non-200 response when trying to connect, or
    // if the max number of retries on network errors is reached.

    // ... handle fatal error
  }
);
```

The driver will take care of the initial request to convert to a stream if you provide a Query

```javascript
import { Client, fql } from "fauna"
const client = new Client({ secret: FAUNA_SECRET })

const stream = await client.stream(fql`MyCollection.all().toStream()`)

  // ...
}
```

### Error handling

The driver tries to reestablish streaming connections when there are network errors. With current defaults, retries start quickly but have exponential backoff, capped at 10s, and will continue trying to reestablish for about 10 minutes. That's generous enough to let your internet connection drop and come back.

Network errors are hidden from the user while retries are still being attempted
 
Errors from Fauna are assumed fatal. Any error events coming from Fauna will cause an error to be thrown.

## Out of scope
N/A

## Testing
tests have been added and can be run on the most recent docker image using `yarn test`.

----

### DIFF
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -156,6 +156,24 @@ export interface Endpoints {
 }
 
 /**
+ * Configuration for a streaming client. This typically comes from the `Client`
+ * instance configuration.
+ */
+export type StreamClientConfiguration = {
+  /**
+   * A secret for your Fauna DB, used to authorize your queries.
+   * @see https://docs.fauna.com/fauna/current/security/keys
+   */
+  secret: string;
+
+  /**
+   * Controls what Javascript type to deserialize {@link https://fqlx-beta--fauna-docs.netlify.app/fqlx/beta/reference/language/types#long | Fauna longs} to.
+   * @see {@link ClientConfiguration.long_type}
+   */
+  long_type: "number" | "bigint";
+};
+
+/**
  * A extensible set of endpoints for calling Fauna.
  * @remarks Most clients will will not need to extend this set.
  * @example
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -171,6 +171,24 @@ export type StreamClientConfiguration = {
    * @see {@link ClientConfiguration.long_type}
    */
   long_type: "number" | "bigint";
+
+  /**
+   * Max attempts for retryable exceptions.
+   */
+  max_attempts: number;
+
+  /**
+   * Max backoff between retries.
+   */
+  max_backoff: number;
+
+  /**
+   * Indicates if stream should include "status" events, periodic events that
+   * update the client with the latest valid timestamp (in the event of a
+   * dropped connection) as well as metrics about about the cost of maintaining
+   * the stream other than the cost of the received events.
+   */
+  status_events?: boolean;
 };
 
 /**
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -1,3 +1,4 @@
+import { HTTPStreamClient } from "./http-client";
 import type { ValueFormat } from "./wire-protocol";
 
 /**
@@ -161,10 +162,9 @@ export interface Endpoints {
  */
 export type StreamClientConfiguration = {
   /**
-   * A secret for your Fauna DB, used to authorize your queries.
-   * @see https://docs.fauna.com/fauna/current/security/keys
+   * The underlying {@link HTTPStreamClient} that will execute the actual HTTP calls
    */
-  secret: string;
+  httpStreamClient: HTTPStreamClient;
 
   /**
    * Controls what Javascript type to deserialize {@link https://fqlx-beta--fauna-docs.netlify.app/fqlx/beta/reference/language/types#long | Fauna longs} to.
@@ -181,6 +181,12 @@ export type StreamClientConfiguration = {
    * Max backoff between retries.
    */
   max_backoff: number;
+
+  /**
+   * A secret for your Fauna DB, used to authorize your queries.
+   * @see https://docs.fauna.com/fauna/current/security/keys
+   */
+  secret: string;
 
   /**
    * Indicates if stream should include "status" events, periodic events that
--- a/src/client.ts
+++ b/src/client.ts
@@ -20,6 +20,7 @@ import {
   ThrottlingError,
   ContendedTransactionError,
   InvalidRequestError,
+  StreamError,
 } from "./errors";
 import {
   HTTPStreamClient,
@@ -326,16 +327,17 @@ export class Client {
     const backoffMs =
       Math.min(Math.random() * 2 ** attempt, maxBackoff) * 1_000;
 
-    const wait = (ms: number) => new Promise((r) => setTimeout(r, ms));
     attempt += 1;
-    return this.#query<T>(queryInterpolation, options, attempt).catch((e) => {
-      if (e instanceof ThrottlingError && attempt < maxAttempts) {
-        return wait(backoffMs).then(() =>
-          this.#queryWithRetries<T>(queryInterpolation, options, attempt)
-        );
+
+    try {
+      return await this.#query<T>(queryInterpolation, options, attempt);
+    } catch (error) {
+      if (error instanceof ThrottlingError && attempt < maxAttempts) {
+        await wait(backoffMs);
+        return this.#queryWithRetries<T>(queryInterpolation, options, attempt);
       }
-      throw e;
-    });
+      throw error;
+    }
   }
 
   #getError(e: any): ClientError | NetworkError | ProtocolError | ServiceError {
@@ -621,6 +623,7 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
 export type StreamEventHandler = (event: StreamEvent) => void;
 
 export class StreamClient {
+  closed = false;
   #callbacks: Record<StreamEventType, StreamEventHandler[]> = {
     start: [],
     add: [],
@@ -631,7 +634,10 @@ export class StreamClient {
   #query: () => Promise<StreamToken>;
   #clientConfiguration: Record<string, any>;
   #httpStreamClient: HTTPStreamClient;
-  #streamAdapter: StreamAdapter | null = null;
+  #connectionAttempts = 0;
+  #streamAdapter?: StreamAdapter;
+  #streamToken?: StreamToken;
+  #last_ts?: number;
 
   // TODO: make clientConfiguration and httpStreamClient optional
   constructor(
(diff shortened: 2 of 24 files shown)


src/http-client/http-client.ts
```diff
@@ -65,3 +66,65 @@ export interface HTTPClient {
...
+//    */
+//   // TODO: return `AsyncGenerator<StreamEvent>`?
+//   start: () => void;
+
+//   /**
+//    * Close the stream.
+//    */
+//   close: () => void;
+// }
+
```
ptpaterson, line 105: oops

```suggestion
```

src/client.ts
```diff
@@ -569,6 +600,102 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+    query: StreamToken,
+    clientConfiguration: StreamClientConfiguration,
+    httpStreamClient: HTTPStreamClient
+  ) {
+    this.#query = query;
+    this.#clientConfiguration = clientConfiguration;
+    this.#httpStreamClient = httpStreamClient;
+  }
+
+  on(type: StreamEventType, callback: StreamEventHandler) {
```
marrony, line 629: we probably should give the option to unsubscribe an existing handler

src/client-configuration.ts
```diff
@@ -155,6 +155,24 @@ export interface Endpoints {
...
+ */
+export type StreamClientConfiguration = {
+  /**
+   * A secret for your Fauna DB, used to authorize your queries.
+   * @see https://docs.fauna.com/fauna/current/security/keys
+   */
+  secret: string;
+
+  /**
+   * Controls what Javascript type to deserialize {@link https://fqlx-beta--fauna-docs.netlify.app/fqlx/beta/reference/language/types#long | Fauna longs} to.
```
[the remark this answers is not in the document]
ptpaterson: Ah. Looks like all of the links should be updated!

src/client.ts
```diff
@@ -569,6 +600,102 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+    for await (const event of streamAdapter.read) {
+      // stream events are always tagged
+      const deserializedEvent: StreamEvent = TaggedTypeFormat.decode(event, {
+        long_type: this.#clientConfiguration.long_type,
+      });
+
+      // TODO: handle callbacks when using the async generator?
+      // this.#callbacks[deserializedEvent.type].forEach((callback) =>
+      //   callback(deserializedEvent)
+      // );
```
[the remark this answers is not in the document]
ptpaterson: Yes, definitely deserves more discussion.

If we drop the async iterator API, it will make the http2 implementation much easier. the http2 module uses callbacks itself, and that is proving to be a pain to use those to provide an iterator interface.

src/http-client/fetch-client.ts
```diff
@@ -64,8 +69,88 @@ export class FetchClient implements HTTPClient {
...
 
+  /** {@inheritDoc HTTPStreamClient.stream} */
+  stream({
+    data,
+    headers: requestHeaders,
+    method,
+  }: HTTPStreamRequest): StreamAdapter {
+    const request = new Request(this.#streamURL, {
+      method,
+      headers: { ...requestHeaders, "Content-Type": "application/json" },
```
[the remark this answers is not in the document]
ptpaterson: That's why the driver does `body: JSON.stringify(data)`.

The API will also eventually support accepting a Query object as JSON, correct? It seems like it would be better to not have to flip the content type depending on the input.

src/http-client/fetch-client.ts
```diff
@@ -64,8 +69,88 @@ export class FetchClient implements HTTPClient {
...
 }
+
+/**
+ * Get individual lines from the stream
+ *
+ * The stream may be broken into arbitrary chunks, but the events are delimited by a newline character.
+ *
+ * @param reader - The stream reader
+ */
+async function* readLines(reader: ReadableStreamDefaultReader<Uint8Array>) {
```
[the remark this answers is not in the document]
ptpaterson: A string in the response shouldn't ever include an actual `\n` character. I expect it would just be a literally the characters '\ and 'n'. But it would be good to check explicitly, and ultimately include in some test automation once we can use this in Node. 👍🏻

src/http-client/http-client.ts
```diff
@@ -1,5 +1,6 @@
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
 import type { Client } from "../client";
+// import type { Stream } from "../stream";
```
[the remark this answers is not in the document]
ptpaterson: yup!

src/client.ts
ptpaterson: That would be nice. Do you have thoughts on what would be a good way to do that? Note that this implementation allows multiple callbacks for the same event type.

A few ways come to mind for me:

1. `remove(type: StreamEventType)`, removes ALL callbacks for a given event type.
2. Provide a third parameter for options when adding a callback, which could include a name/id for the callback. e.g. `on(type: StreamEventType, callback: StreamEventHandler), options: { id: string })`. Then you could do `stream.remove('callbackName').
3. return a handler to the callback from `on`, instead of the StreamClient. This would mean that you couldn't chain `.on(...)` calls, but you could stash that callback to cancel it later.

```
// option 3
const stream = client.stream(token)
const addHandle = stream.on('add', (e) => ...)

// ... later
addHandle.remove() // not sure what to do with this dangling handler now, though.
```

erickpintor: It seems a bit cumbersome to remove a single handler after the stream starts. It might error prone to setup the client, then start the stream, then at some point later a callback stops being called because some other part of the code base removed a handler. If we need to support advanced event handling, I'd prefer a more general event callback where you catch all events instead. Users can write their own complex logic there.

erickpintor: At a second look, it might be even confusing to users that we design an API for selecting event based on their types but, they get all events in the wire! One might expect that they are saving on network footprint by only handling some events. Perhaps the more general "catch all" API would make the whole iteration a bit more clear overall.

ptpaterson: > Perhaps the more general "catch all" API would make the whole iteration a bit more clear overall.

Do you mean the async-iterator API? e.g. `for await (const event in stream.iter())`?

src/http-client/fetch-client.ts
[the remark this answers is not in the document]
erickpintor: > That's why the driver does body: JSON.stringify(data).

The `stringify` is not called for the streaming endpoint. I'm sure you'd be betting a 400 if you were sending the token with quotes in it.

> The API will also eventually support accepting a Query object as JSON, correct? It seems like it would be better to not have to flip the content type depending on the input.

I'm not sure when / if that will came to be, honestly. In any case, when that happens, it'll require driver work to wire things up. We may switch to a more fleshed out JSON request body then.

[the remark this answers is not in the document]
ptpaterson: I'll set content type to text/plain for now.

src/client.ts
erickpintor: Idk which API shape is the most idiomatic. My point is: having the driver select event types may confuse users into thinking they are only getting those events.

The async-iterator API as is right now aligns with my perspective since it does clearly informs users that they are getting all events. The callback based API would have to be reshaped into `.on(event => {..})` to align with what I'm describing here.

src/client.ts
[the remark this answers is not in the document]
fauna-chase: personal opinion here, but my preference would be to just go with the callback api for now.  That callback pattern feels fairly common in js.  I agree that having one method  of doing it seems like a simpler and more straight forward  starting point.  
Could be good to get some input from some of th dx folks who have a bit more experience on the javascript side.

src/client.ts
fauna-chase: feel free to correct me if I'm wrong but I think he is referring to something like the following:
```
const stream = client.stream(stream)
      .onEvent((type, data) => {
          switch (type) {
          case 'update':
              console.log(`New doc's data ${data}`)
              break
          case 'remove':
              console.log('Doc deleted')
              break
          }
      })
      .onError(error => {
          console.log(error)
      })
```
though I suppose in that example we had the error case special handled.  could move that into the general event handling as well though I don't think it is necessarily the worst to handle errors in that way.

src/client.ts
```diff
@@ -569,6 +600,102 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+        });
+        this.#callbacks[deserializedEvent.type].forEach((callback) =>
+          callback(deserializedEvent)
+        );
+      }
+    };
+
+    handleEvents().catch((error) => {
+      throw new NetworkError("Error reading from stream", { cause: error });
+    });
```
fauna-chase, line 661: how would this trigger in the user application code? where would I add my error handling for this scenario?

fauna-chase (commented):
one thing I wonder, and we could always wait and see if people want this, but would it be worth in addition to our two step
1 - create stream token
2 - call stream

to also add an API in the drivers that allowed
client.stream(fql`query here that produces a stream`).on( event => ..)
purely as a form of convenience.  Under the hood we'd still send the same queries.

src/client.ts
ptpaterson: You have to put `stream.start()` or `for await (event of stream.iter())` in a try/catch block.

src/client.ts
ptpaterson: I don't think it's a problem that all events are being triggered even if you don't register a callback for them. This should be familiar to JS devs. This is how v4 driver works for reference, not that that means we _should_ do it that way.

Should registering a callback overwrite the existing callback, or should it be added to a list?

If we break up the `on` method to have multiple such as `onStart`, `onEvent`/`onData` and `onError` methods, then how do you want to do de-registering events?

ptpaterson: FWIW, if we are going to only have one method between callbacks and async iterator, then my vote is currently on async iterator. Then this becomes moot, and, to Chase's question on error handling, it is clear where any network errors would be caught and handled.

src/client.ts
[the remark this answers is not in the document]
ptpaterson: my vote is currently on async iterator if we only go with one interface. I think it eliminates a lot of subjective questions about how to register/de-register callbacks and how to handle thrown network errors.

src/client.ts
ptpaterson: Actually I don't think that this Network error would be caught by a try catch block around `stream.start()` and the callback method. I think we actually would need to pipe this into the error callback. And then I don't know what to do if the user doesn't register one. That's a problem that also exists with the v4 driver.

ptpaterson:
> > one thing I wonder, and we could always wait and see if people want this, but would it be worth in addition to our two step 1 - create stream token 2 - call stream
> > to also add an API in the drivers that allowed `client.stream(fql'query here that produces a stream').on( event => ..)` purely as a form of convenience. Under the hood we'd still send the same queries.
> 
> I think it's a good idea.

This will turn the `Client.stream` method into an async method.

ptpaterson:
> > one thing I wonder, and we could always wait and see if people want this, but would it be worth in addition to our two step 1 - create stream token 2 - call stream
> > to also add an API in the drivers that allowed `client.stream(fql'query here that produces a stream').on( event => ..)` purely as a form of convenience. Under the hood we'd still send the same queries.
> 
> I think it's a good idea.

I've implemented this. It required Client.stream to be async, but I think that is okay. It will only make the initial request for the stream token. The stream connection still waits for you to call stream.start or start iterating.

src/client.ts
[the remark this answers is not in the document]
ptpaterson: I confirmed that network errors are not caught when using the callback API. We would need to pipe any errors into the error handler.

src/http-client/fetch-client.ts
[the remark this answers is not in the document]
ptpaterson: Body was updated to an object to handle start timestamps.

```typescript
export type StreamRequest = {
  token: string;
  start_ts?: number;
};
```

ptpaterson:
I've updated error handling, so that retry-able network errors create events which can be handled. In the event of a retry, the user should expect at least two events back-to-back:
1. One or more "error" events with info about the network error(s), and
2. assuming a new connection is made, another "start" event

src/client.ts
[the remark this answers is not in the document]
ptpaterson: Any errors are now caught and handled by an optional handler.

ptpaterson:
@pnwpedro I've rebased on main to include the concourse pipeline changes, and this PR is now set to merge into the `beta` branch.

pnwpedro (approved):

PR #235, opened 2024-03-08
Initial implementation for streams with http2 module

ptpaterson:
Ticket(s): FE-4969

## Problem
The first iteration of v10 streaming did provide a default implementation for Node.js

## Solution
Add a streaming implementation to the `NodeHTTP2Client` class

## Result
V10 streams can be used in Node.js by default

## Out of scope
There are changes to the error shape on the core side that might not initially be incorporated here.

## Testing
I am running tests with a Node.js script running locally on my machine.

----

### DIFF
--- a/src/http-client/node-http2-client.ts
+++ b/src/http-client/node-http2-client.ts
@@ -247,42 +247,44 @@ export class NodeHTTP2Client implements HTTPClient, HTTPStreamClient {
         http2ResponseHeaders[http2.constants.HTTP2_HEADER_STATUS]
       );
       if (!(status >= 200 && status < 400)) {
-        // TODO: if we get bad status, then we should still finish reading the response, and create
-        // the appropriate error instance
-        rejectChunk(
-          new ServiceError(
-            {
-              error: {
-                code: "fauna error",
-                message: "fauna error",
-              },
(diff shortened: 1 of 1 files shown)


src/http-client/node-http2-client.ts
```diff
@@ -213,4 +221,107 @@ export class NodeHTTP2Client implements HTTPClient {
...
+    let req: ClientHttp2Stream;
+    const onResponse = (
+      http2ResponseHeaders: IncomingHttpHeaders & IncomingHttpStatusHeader
+    ) => {
+      const status = Number(
+        http2ResponseHeaders[http2.constants.HTTP2_HEADER_STATUS]
+      );
+      if (!(status >= 200 && status < 400)) {
+        // TODO: if we get bad status, then we should still finish reading the response, and create
+        // the appropriate error instance
```
ptpaterson, line 251: We need to branch here to finish reading the data and throw an appropriate Fauna error, instead of trying to yield any events to the stream.

ptpaterson: The branching logic is done. As a follow-up, we should let the Client do the work to identify what kind of error was received.

PR #236, opened 2024-03-21
Setup pipeline for beta deploy

pnwpedro:
Ticket(s): FE-###

## Problem
We don't have a workflow that supports releasing beta versions of the driver without merging into main.

## Solution
Create a beta branch (protected) and extend the pipeline to support releases from that branch.

## Testing
The pipeline has been set. Internal reviewers can find it.

----


ptpaterson (approved):
LGTM

PR #252, opened 2024-04-02
Sort service errors into the proper subclass

ptpaterson:
Ticket(s): [FE-5040](https://faunadb.atlassian.net/browse/FE-5040)

## Problem
`HTTPStreamClient` implementations need to throw service errors if the initial response is a non-200. The logic to convert `QueryFailure` responses to the appropriate sub type is currently a private method on the `Client` class.

The `StreamClient` also needs to throw the appropriate service error class when a type=error event is received.

## Solution
Move the `Client.#getServiceError` method to the `errors` module so it can be reused. Use it to throw the appropriate error if a type=error event is received or if the initial response is a non-200.

There is no http status code associated with streaming error events, so we should update error handling to rely first on the error code, then fallback to the base `ServiceError` if the error code is not matched.

## Out of scope
N/A

## Testing
Updated tests to check that the appropriate errors are thrown.

----

[FE-5040]: https://faunadb.atlassian.net/browse/FE-5040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/integration/stream.test.ts
+++ b/__tests__/integration/stream.test.ts
@@ -247,7 +247,7 @@ describe("StreamClient", () => {
     }
   });
 
-  it("handles InvalidRequestError when via callback when establishing a stream", async () => {
+  it("handles InvalidRequestError via callback when establishing a stream", async () => {
     expect.assertions(1);
 
     // create a stream with a bad token
--- a/__tests__/integration/stream.test.ts
+++ b/__tests__/integration/stream.test.ts
@@ -289,7 +289,7 @@ describe("StreamClient", () => {
       }
     } catch (e: any) {
       expect(e).toBeInstanceOf(AbortError);
-      expect(e.httpStatus).toBe(-1);
+      expect(e.httpStatus).toBeUndefined();
       expect(e.abort).toBe("oops");
     } finally {
       stream?.close();
@@ -316,7 +316,7 @@ describe("StreamClient", () => {
       }
     } catch (e: any) {
       expect(e).toBeInstanceOf(QueryRuntimeError);
-      expect(e.httpStatus).toBe(-1);
+      expect(e.httpStatus).toBeUndefined();
     } finally {
       stream?.close();
     }
@@ -347,7 +347,7 @@ describe("StreamClient", () => {
       function onEvent(_) {},
       function onError(e) {
         if (e instanceof AbortError) {
-          expect(e.httpStatus).toBe(-1);
+          expect(e.httpStatus).toBeUndefined();
           expect(e.abort).toBe("oops");
         }
         resolve();
@@ -382,7 +382,7 @@ describe("StreamClient", () => {
       function onEvent(_) {},
       function onError(e) {
         if (e instanceof QueryRuntimeError) {
-          expect(e.httpStatus).toBe(-1);
+          expect(e.httpStatus).toBeUndefined();
         }
         resolve();
       },
--- a/__tests__/integration/stream.test.ts
+++ b/__tests__/integration/stream.test.ts
@@ -1,15 +1,16 @@
 import {
-  fql,
-  getDefaultHTTPClient,
+  AbortError,
+  Client,
+  DocumentT,
+  DateStub,
+  Document,
+  InvalidRequestError,
   StreamClient,
   StreamClientConfiguration,
   StreamToken,
-  Client,
-  DocumentT,
-  ServiceError,
   TimeStub,
-  DateStub,
-  Document,
+  fql,
+  getDefaultHTTPClient,
 } from "../../src";
 import {
   getClient,
@@ -19,7 +20,6 @@ import {
 
 const defaultHttpClient = getDefaultHTTPClient(getDefaultHTTPClientOptions());
 const { secret } = getDefaultSecretAndEndpoint();
-const dummyStreamToken = new StreamToken("dummy");
 
 let client: Client;
 const STREAM_DB_NAME = "StreamTestDB";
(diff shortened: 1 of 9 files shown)


__tests__/integration/stream.test.ts
```diff
@@ -320,15 +346,53 @@ describe("StreamClient", () => {
...
+          expect(e.httpStatus).toBe(400);
+        }
         resolve();
       }
     );
 
     await promise;
   });
 
+  // TODO: test other error events
```
cleve-fauna, line 394: is this something we can just do now?

__tests__/integration/stream.test.ts
```diff
@@ -242,12 +243,11 @@ describe("StreamClient", () => {
...
       }
     } catch (e) {
-      // TODO: be more specific about the error and split into multiple tests
-      expect(e).toBeInstanceOf(ServiceError);
+      expect(e).toBeInstanceOf(InvalidRequestError);
     }
   });
 
-  it("handles non 200 responses via callback when establishing a stream", async () => {
+  it("handles InvalidRequestError when via callback when establishing a stream", async () => {
```
cleve-fauna, line 250:
```suggestion
  it("handles InvalidRequestError via callback when establishing a stream", async () => {
```

__tests__/integration/stream.test.ts
ptpaterson: If you can think of ways to trigger them. I know in theory you could, for example, get an error due to throughput limits, but I'm not sure how to reliably test that.

cleve-fauna: Ah i see.

Mocking would be the best bet for those.

src/errors.ts
```diff
@@ -312,3 +312,49 @@ export class ProtocolError extends FaunaError {
...
+
+const QUERY_CHECK_FAILURE_CODES = [
+  "invalid_function_definition",
+  "invalid_identifier",
+  "invalid_query",
+  "invalid_syntax",
+  "invalid_type",
+];
+
+export const getServiceError = (
```
cleve-fauna, line 324: Let's write a unit test for this since it is now a public method.

src/http-client/node-http2-client.ts
```diff
@@ -260,7 +261,16 @@ export class NodeHTTP2Client implements HTTPClient, HTTPStreamClient {
         // TODO: The Client contains the information for how to parse an error
         // into the appropriate class, so lift this logic out of the HTTPClient.
```
cleve-fauna, line 262: can we delete this TODO now?

src/errors.ts
ptpaterson: Good idea! Will do.

cleve-fauna:
@ptpaterson you've got some tests failing.

__tests__/integration/stream.test.ts
cleve-fauna: Try the 2nd approach then with the verificaiton and associated unit test.

ptpaterson:
It's unrelated to the streaming tests. Looks like a check for client-timeout failed.

cleve-fauna:
> It's unrelated to the streaming tests. Looks like a check for client-timeout is being flaky.

Can you try to deflake it?

__tests__/unit/error.test.ts
```diff
@@ -0,0 +1,88 @@
...
+    ${400}     | ${"invalid_receiver"}                   | ${QueryRuntimeError}
+    ${400}     | ${"invalid_timestamp_field_access"}     | ${QueryRuntimeError}
+    ${400}     | ${"invalid_request"}                    | ${InvalidRequestError}
+    ${400}     | ${"abort"}                              | ${AbortError}
+    ${401}     | ${"unauthorized"}                       | ${AuthenticationError}
+    ${403}     | ${"forbidden"}                          | ${AuthorizationError}
+    ${409}     | ${"contended_transaction"}              | ${ContendedTransactionError}
+    ${429}     | ${"throttle"}                           | ${ThrottlingError}
+    ${440}     | ${"time_out"}                           | ${QueryTimeoutError}
+    ${503}     | ${"time_out"}                           | ${ServiceTimeoutError}
```
cleve-fauna, line 66: are these really the same code?

src/errors.ts
```diff
@@ -196,19 +224,24 @@ export class ThrottlingError extends ServiceError {
...
+ * This error can have one of two sources:
+ *     1. Fauna is behaving expectedly, but the query timeout provided was too
+ *        aggressive and lower than the query's expected processing time.
+ *     2. Fauna was not available to service the request before the timeout was
+ *        reached.
+ *
+ * In either case, consider increasing the `query_timeout_ms` configuration for
+ * your client.
  */
 export class QueryTimeoutError extends ServiceError {
```
ptpaterson, line 236: I merged the ServiceTimeoutError class into QueryTimeoutError. We talk about "client timeouts" vs "query timeouts" so I think that's the appropriate name, and QueryTimeoutError already had optional stats field defined.

ptpaterson (commented):
Error classes have been updated.

Fixes:
- `invalid_query` is the only real `QueryCheckError` code

Streaming changes
- `FetchClient` and `NodeHTTP2Client` return `NetworkErrors` when problems are encountered after the stream is connected.

New breaking changes compared to v1.3.1
- The `httpStatus` field for `ServiceError`s is now optional
- Added a `ConstraintFailureError` class and broke that out from `QueryRuntimeError`
- Merged the `ServiceTimeoutError` class into the `QueryTimeoutError` class. They share the same `time_out` code.

cleve-fauna (approved):

PR #260, opened 2024-05-06
Add encoding/decoding for Fauna Bytes

ptpaterson:
Ticket(s): [FE-4855](https://faunadb.atlassian.net/browse/BT-4855)

## Problem
Fauna supports a Bytes type which is transmitted as a base-64-encoded string
https://docs.fauna.com/fauna/current/reference/fql_reference/types#bytes

We also had a type issue where JS Date could not be used as a query argument, so I added Date to arg type as well as a test.

## Solution
Add encoding and decoding of the Bytes type

## Result
`ArrayBuffer`s, `TypedArray`s (e.g. Uint8Array, Int32Array, etc.), as well as Node's `Buffer` classes are encoded to Fauna Bytes.

Fauna Bytes are decoded into `Uint8Array`, which can be used to create the desired type.

See tests for examples.

## Out of scope
N/A

## Testing
Added unit and integration tests for Node. Browser is untested.

----

[FE-4855]: https://faunadb.atlassian.net/browse/FE-4855?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,8 +58,5 @@
     "suiteNameTemplate": "{filepath}",
     "classNameTemplate": "{classname}",
     "titleTemplate": "{title}"
-  },
-  "dependencies": {
-    "base64-js": "^1.5.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -58,5 +58,8 @@
     "suiteNameTemplate": "{filepath}",
     "classNameTemplate": "{classname}",
     "titleTemplate": "{title}"
+  },
+  "dependencies": {
+    "base64-js": "^1.5.1"
   }
 }
--- a/src/tagged-type.ts
+++ b/src/tagged-type.ts
@@ -218,7 +218,7 @@ const encodeMap = {
   // TODO: encode as a tagged value if provided as a query arg?
   // streamToken: (value: StreamToken): TaggedStreamToken => ({ "@stream": value.token }),
   streamToken: (value: StreamToken): string => value.token,
-  bytes: (value: Uint8Array): TaggedBytes => ({
+  bytes: (value: ArrayBuffer | ArrayBufferView): TaggedBytes => ({
     "@bytes": bufferToBase64(value),
   }),
 };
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -58,5 +58,8 @@
     "suiteNameTemplate": "{filepath}",
     "classNameTemplate": "{classname}",
     "titleTemplate": "{title}"
+  },
+  "dependencies": {
+    "base64-js": "^1.5.1"
   }
```
cleve-fauna, line 64: hmmm, can we avoid this?

Right now we have **no dependencies** which is awesome.

Can we not stick with native code?

ptpaterson: Removed the dep. We now check for Node and use the `Buffer` class to do the encoding/decoding. Use browser provided `btoa` and `atob` if not node.

src/tagged-type.ts
```diff
@@ -261,9 +269,29 @@ const encode = (input: QueryValue): QueryValue => {
...
+      } else if (input instanceof ArrayBuffer || ArrayBuffer.isView(input)) {
+        return encodeMap["bytes"](input);
       } else {
         return encodeMap["object"](input);
       }
   }
   // anything here would be unreachable code
 };
+
+function base64toBuffer(value: string): Uint8Array {
```
freels, line 281: Shouldn't this return an ArrayBuffer?

ptpaterson: I switched the decoding output to be Uint8Array based on feedback. We can rename these helper functions to "...Array..." instead of "...Buffer..." if that helps.

freels: If we decode to Uint8Array I do not think we should not accept any other TypedArray variant on encode. It's too easy to get send one thing and get a slightly different thing back. So accept only ArrayBuffers and Uint8Arrays.

ptpaterson: Done in [e06dabc](https://github.com/fauna/fauna-js/pull/260/commits/e06dabc0432a3d1cd6c0e17376897286ba6c558d)

freels (approved):
Take it with a grain of salt, but LGTM

https://i.kym-cdn.com/entries/icons/original/000/008/342/ihave.jpg

PR #272, opened 2024-06-12
Support users embedding fql snippets in objects and arrays

ptpaterson:
Ticket(s): [FE-3501](https://faunadb.atlassian.net/browse/FE-3501)

## Problem
`Query` instances (i.e. results from the `fql` function) cannot be encoded when embedded in objects or arrays.

The `QueryValue` type is incorrectly used for types over the wire.

## Solution
The database now accepts a query interpolation fragments for `array` and `object`, where the existing `fql` fragment can now be nested under the `array` and `object` fragments. Add new `encodeInterpolation` method to encode with the new fragments.

Query `arguments` do not use interpolation, so they still need the `@object` tag. Encoding arguments still uses the existing `encode` method.

Update types to better distinguish between
- Arguments to be encoded
  - allow `Query` within objects and arrays
  - `QueryValue` should remain a valid `QueryArgument` to ensure round-trip-ability
- Encoded values sent over the wire
  - Updated fragment types and remove `QueryValue` from fragments
- Decoded values
  - Specify `QueryValue` type only where we decode values from Fauna.

## Result
Queries can be encoded in QueryInterpolations.

Queries can be embedded within JS arrays and objects.

Type changes are corrections to the internal implementation. No user-facing changes were made to types (except `QueryArgument`)

## Out of scope
n/a

## Testing
Updated existing tests and added new ones for embedding queries in arrays and objects.

----

[FE-3501]: https://faunadb.atlassian.net/browse/FE-3501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/query-builder.ts
+++ b/src/query-builder.ts
@@ -75,19 +75,16 @@ export class Query {
   }
 
   /**
-   * Converts this Query to a {@link QueryRequest} you can send
+   * Converts this Query to an {@link FQLFragment} you can send
    * to Fauna.
-   * @param options - optional {@link QueryOptions} to include
-   *   in the request (and thus override the defaults in your {@link ClientConfiguration}.
-   *   If not passed in, no headers will be set as overrides.
-   * @returns a {@link QueryRequest}.
+   * @returns a {@link FQLFragment}.
    * @example
    * ```typescript
    *  const num = 8;
    *  const queryBuilder = fql`'foo'.length == ${num}`;
    *  const queryRequest = queryBuilder.toQuery();
    *  // produces:
-   *  { query: { fql: ["'foo'.length == ", { value: { "@int": "8" } }, ""] }}
+   *  { fql: ["'foo'.length == ", { value: { "@int": "8" } }, ""] }
    * ```
    */
   encode(): FQLFragment {
--- a/src/query-builder.ts
+++ b/src/query-builder.ts
@@ -8,16 +8,17 @@ import type {
 /**
  * A QueryArgumentObject is a plain javascript object where each property is a
  * valid QueryArgument.
- * These objects can be set as values in the {@link fql} function.
  */
 export type QueryArgumentObject = {
   [key: string]: QueryArgument;
 };
 
 /**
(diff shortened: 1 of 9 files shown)


src/query-builder.ts
```diff
@@ -1,18 +1,32 @@
...
+ * valid QueryArgument.
+ * These objects can be set as values in the {@link fql} function.
+ */
+export type QueryArgumentObject = {
+  [key: string]: QueryArgument;
+};
+
+/**
+ * A QueryArgument is a plain javascript object where each property is a
+ * valid QueryArgument.
```
pnwpedro, line 19: I don't think this comment is correct

pnwpedro (changes requested):
One small fix to a comment please. Looks good otherwise

src/tagged-type.ts
```diff
@@ -14,7 +14,26 @@ import {
...
+  TaggedDouble,
+  TaggedObject,
+  EncodedObject,
+  TaggedTime,
+  TaggedDate,
+  TaggedMod,
+  TaggedRef,
+  TaggedBytes,
+} from "./wire-protocol";
+import { Query, QueryArgument, QueryArgumentObject } from "./query-builder";
```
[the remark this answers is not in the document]
ptpaterson: I had some reorganization of types staged but I put everything back. I was worried about 
1. breaking import paths, and
2. complicating the diff for this PR.

On further consideration of no. 1, that doesn't actually matter because we bundle the driver and only expose the exports from src/index and that's it.

So I guess my point is: agreed! We should come back to this.

src/query-builder.ts
ptpaterson: Thanks for catching the copy-pasta! I've updated in [7bb285a](https://github.com/fauna/fauna-js/pull/272/commits/7bb285a318d9096ded589200a4e2ee7bab325e81). How's that?

pnwpedro (approved):

PR #275, opened 2024-07-17
Fix incorrect paths type on ConstraintFailure

ecooper:
Ticket(s): [FE-5552](https://faunadb.atlassian.net/browse/FE-5552)

## Problem

The type `ConstraintFailure["paths"]` was set to be `Array<number | string>`. However, the wire response for `paths` is actually an array of arrays. Anyone trying to interact with `paths` using the declared types would have a runtime error. The existing integration tests only check `paths` is defined.

## Solution

Fix the the types to match the current response, and update the integration tests to actually test the array of array data structure. The `expect` statements were updated to use more "complex" matchers to reduce the number of assertions while visually documenting the data structures.

## Result

The types now match the expected wire result and the integration tests validate that structure more concretely.

## Out of scope

N/A.

## Testing

Updated the existing constraint failure query tests.

----

[FE-5552]: https://faunadb.atlassian.net/browse/FE-5552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/integration/query.test.ts
+++ b/__tests__/integration/query.test.ts
@@ -229,7 +229,7 @@ describe("query", () => {
   });
 
   it("Includes constraint failures when present", async () => {
-    expect.assertions(6);
+    expect.assertions(4);
     try {
       await client.query(
         fql`Function.create({"name": "my_double", "body": "x => x * 2"})`,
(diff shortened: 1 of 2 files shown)


__tests__/integration/query.test.ts
```diff
@@ -241,14 +241,15 @@ describe("query", () => {
...
-            expect(constraintFailure.paths).toBeDefined();
-          }
-        }
+        expect(e.queryInfo).toMatchObject({
+          summary: expect.any(String),
+        });
+        expect(e.constraint_failures).toEqual([
+          {
+            message: expect.any(String),
+            paths: expect.arrayContaining([expect.any(Array)]),
```
ptpaterson, line 250: I couldn't figure out a simple way to assert the array contains `String | Number`. I guess we could extend `expect` but doesn't seem worth it. This is much improved already.

ptpaterson (approved):
LGTM

PR #277, opened 2024-07-22
Add backwards compatible generics support for fql statements

ecooper:
Ticket(s): [FE-5389](https://faunadb.atlassian.net/browse/FE-5389)

## Problem

Today, fql statements return a `Query` type. We'd like for fql statements to carry generic types that describe the resulting value once it's evaluated. This would allow users to "type" a fql statement once and have it be inferred in client methods.  This would eliminate the need to type client call methods.

We want to do this in a backwards compatible way.

## Solution

By adding a type parameter that extends `QueryValue` to fql statements and `Query`, we can enable types to be inferred in client method calls with minimal changes. So something like this "just works":

```ts
const query = fql<User>`{
  name: "Alice",
  email: "alice@site.example",
}`;

// response will be typed as QuerySuccess<User>
const response = await client.query(query);

// userDoc will be automatically inferred as User
const userDoc = response.data;
```

## Result

The following PR adds support for generics in fql statements and `Query` in a backwards compatible way. The following usages work:

```ts
// Use fql to type client return values...
const query = fql<User>`{
  name: "Alice",
  email: "alice@site.example",
}`;

// response will be typed as QuerySuccess<User>
const response = await client.query(query);

// userDoc will be automatically inferred as User
const userDoc = response.data;
```

Similarly, you can still type client calls directly:

```ts
// Don't type fql
const query = fql`{
  name: "Alice",
  email: "alice@site.example",
}`;

// response will be typed as QuerySuccess<User>
const response = await client.query<User>(query);

// userDoc will be User
const userDoc = response.data;
```

Thanks to the efforts of @ptpaterson, users are protected from conflicting types when they attempt to type _both_ fql and client methods. See the following test for this in action:

```ts
  it("allows customers to use subtyped queries", async () => {
    const query = fql<string>`"hello"`;

    const result = (await client.query<string | number>(query)).data;
    expect(result).toEqual("hello");

    // And make sure that the opposite is not possible
    const query2 = fql<string | number>`"hello"`;
    // @ts-expect-error Argument of type 'Query<string | number>' is not assignable to parameter of type 'Query<string>'.
    await client.query<string>(query2);
```

## Out of scope

* Backwards incompatible changes
* Ability to compose types across multiple fql statements
* Large-scale type rewrites--especially since `Query` is publicly exported

## Testing

To test, the existing query-typing tests were expanded to demonstrate fql types.

----

[FE-5389]: https://faunadb.atlassian.net/browse/FE-5389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


__tests__/integration/query-typings.test.ts
```diff
@@ -27,4 +28,28 @@ describe.each`
...
+    const query = fql<MyType>`{ "x": 123 }`;
+    const q2 = fql`{ "x": ${query} }`;
+
+    expect.assertions(1);
+    if ("query" === method) {
+      const result = (await client.query(query)).data;
+      noopToValidateInferredType(result);
+      expect(result).toEqual({ x: 123 });
+    } else {
+      for await (const page of client.paginate<MyType>(fql`{ "x": 123}`)) {
```
ptpaterson, line 46: q2 is unused. 

Also, while providing a query that returns a single value is technically allowed, it's not "proper". It's really just a fallback which takes the return value and turns it into an array with a single item: the result of the query.

We should test a query (and type it so) that returns the FQL type `Set`. We should do this in the above test as well.

If we do the following, we actually, currently get a type error, because the current signature of `Client.paginate` accepts a `Query<T>` as argument, but what we (typically) want is the argument to be `Query<Page<T>>`. I guess we need to allow both.

```js
    const query = fql<MyType>`{ "x": 123 }`;
    const paginatedQuery = fql`[{ "x": 123 }].toSet()`;

    expect.assertions(1);
    if ("query" === method) {
      const result = (await client.query(query)).data;
      noopToValidateInferredType(result);
      expect(result).toEqual({ x: 123 });
    } else {
      // type error if `Query<Page<T>>` is not allowed as an argument
      for await (const page of client.paginate<MyType>(paginatedQuery) { 
     
  // ...
```

src/client.ts
```diff
@@ -211,7 +211,7 @@ export class Client {
    * ```
    */
   paginate<T extends QueryValue>(
-    iterable: Page<T> | EmbeddedSet | Query,
+    iterable: Page<T> | EmbeddedSet | Query<T>,
```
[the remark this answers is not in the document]
ptpaterson: Going to need to be `Query<T> | Query<Page<T>>`. See comments for query typing tests.

ptpaterson (changes requested):
The type for `Client.paginate()` needs a tweek. I've a change coming for this.

Also, do we want to include `Client.stream()` in this PR, too?

mwilde345 (approved):

src/client.ts
```diff
@@ -269,9 +270,11 @@ export class Client {
 
   /**
    * Initialize a streaming request to Fauna
+   * @typeParam T - The expected type of the response from Fauna. T can be inferred
+   *   if theprovided query used a type parameter.
```
cleve-fauna, line 274: /s/theprovided/the provided/

__tests__/integration/query-typings.test.ts
```diff
@@ -12,19 +15,67 @@ describe.each`
...
+  it("allows customers to use subtyped queries", async () => {
+    const query = fql<string>`"hello"`;
+
+    const result = (await client.query<string | number>(query)).data;
+    expect(result).toEqual("hello");
+
+    // And make sure that the opposite is not possible
+    const query2 = fql<string | number>`"hello"`;
+    // @ts-expect-error Argument of type 'Query<string | number>' is not assignable to parameter of type 'Query<string>'.
+    await client.query<string>(query2);
```
cleve-fauna, line 79: wow cool assertion here.

cleve-fauna (commented):
Looks good! I like the solution. Found one small doc typoe.

__tests__/integration/query-typings.test.ts
cleve-fauna: Would you mind adding how to use this to the development section of the README?

As we tweak type arguments in future work capturing that we can sanity check our types using this and proper linter settings will help out future devs.

For example, we got some work we'd like to do to make less boilerplate for users when definining their own types and when using `.query` calls that return pages (as opposed to using the clients pagination utils - there's cases you might do that in an API for example). We have some of these type improvements documented internally here: https://faunadb.atlassian.net/wiki/spaces/DX/pages/3852697604/Developer+Tooling+Pain#API-Pagination

So this test pattern is something we will likely need aagain.

PR #281, opened 2024-07-30
Increase query limit integration test timeout value

ecooper:
## Problem

The query limit integration tests can sometimes timeout with at 20000. This makes our release pipeline flaky for new versions. 

## Solution

Increasing the query limit timeout to 30000 should eliminate timeout failures outside of events where dependencies are actually failing.

## Testing

None. This is a theory with minimal risk/downside.

----


mwilde345 (approved):

ptpaterson (approved):
Agreed about little risk/downside to increasing Jest's test timeouts. They cause me problems all the time running the other tests locally.

PR #285, opened 2024-08-21
ENG-6606: Introduce event cursor support

erickpintor:
### Description

Provide support for the new per-event stream cursor.

### Motivation and context

There's an edge case where resuming a stream from a txn time with lots of documents written could result in event loss. This approach fixes that edge case.

Unit tests included.

NA

### Change types

* - [X] Bug fix (non-breaking change that fixes an issue)
* - [X] New feature (non-breaking change that adds functionality)

### Checklist:

* - [X] My code follows the code style of this project.
* - [X] My change requires a change to Fauna documentation.
* - [X] My change requires a change to the README, and I have updated it accordingly.

----

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -344,6 +344,15 @@ export class Client {
         ...options,
       };
 
+      if (
+        streamClientConfig.cursor !== undefined &&
+        tokenOrQuery instanceof Query
+      ) {
+        throw new ClientError(
+          "The `cursor` configuration can only be used with a stream token.",
+        );
+      }
+
       const tokenOrGetToken =
         tokenOrQuery instanceof Query
           ? () => this.query<StreamToken>(tokenOrQuery).then((res) => res.data)
@@ -636,6 +645,8 @@ export class StreamClient<T extends QueryValue = any> {
   #query: () => Promise<StreamToken>;
   /** The last `txn_ts` value received from events */
   #last_ts?: number;
+  /** The last `cursor` value received from events */
+  #last_cursor?: string;
   /** A common interface to operate a stream from any HTTPStreamClient */
   #streamAdapter?: StreamAdapter;
   /** A saved copy of the StreamToken once received */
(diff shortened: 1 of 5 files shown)


src/client.ts
```diff
@@ -636,6 +645,8 @@ export class StreamClient<T extends QueryValue = any> {
   #query: () => Promise<StreamToken>;
   /** The last `txn_ts` value received from events */
   #last_ts?: number;
+  /** The last `cursor` value received from events */
+  #last_cursor?: string;
```
ptpaterson, line 649: `cursor` was added as an option, but this is private so folks won't be able to pick it out of the stream client after an error. I think we can make this easier.

Options:
1. Do nothing: folks can store their own copy of the cursor as events come in, it's just an extra thing they have to remember to do.
2. Add a getter method to read the `#last_cursor`. Then folks could read it from a closed StreamClient and create a new one with the `cursor` option.
3. Remove the `cursor` option and provide a `restart` method that would establish a new connection with the stored `#last_cursor`.

Letting folks configure a cursor option is the most flexible, but if we want to do that, then I think we should make it easier to get the cursor after an error. Having a `restart` method would be one of the easiest things for a customer to use, but maybe not as flexible.

ptpaterson (changes requested):
Nice. This is looking good! Just some thoughts on using the cursor.

src/client.ts
erickpintor: The intent is to be able to recover from a crash, in which case you can't rely on the `last_cursor` as memory would be lost in that case so, I feel like the `restart` would come short in that regard. I'll add a getter since that's pretty harmless but, in the case of crash recovery, users can't rely on that either.

erickpintor: You gotta persist it in durable storage (disk or somewhere else).

pnwpedro: You need to persist it as you handle an event if you want to recover and this impl supports that, so I'm not sure adding a getter is actually harmless. It can throw devs off. They shouldn't implement cursor persistence by polling the client, for example.

ptpaterson: Okay, thanks! I concur with Lucas about not needing the getter.

ptpaterson (approved):

PR #288, opened 2024-10-02
Extend HTTPRequest to support path overrides

ecooper:
This PR extends the existing `HTTPRequest` to take an optional path to support POST requests to endpoints other than `/query/1`.

### Description
We'll need to support requests to Fauna APIs beyond `/query/1`. Today, this API path is hardcoded in both the fetch and http2 client wrappers. This PR does the following:

- Extends `HTTPRequest` to accept an optional `path`
- Updates `FetchClient` to use default API paths, but override per request when one is provided on `HTTPRequest`
- Updates `NodeHTTP2Client` to use default API paths, but override per request when one is provided on `HTTPRequest`
- Adds a readonly `FaunaAPI` to represent the supported API paths in the driver
- Restricts `HTTPRequest["path"]` to the union of supported endpoints via `FaunaAPI`

### Motivation and context
We need to support more endpoints in this driver. The current http clients really only have a fixed concept of requesting queries or streaming. This change allows for a backwards compatible way to introduce providing a path without breaking existing code or interfaces.

Alternatives considered:

- Making `HTTPClient.request` be `(path: SupportedFaunaAPIs, req: HTTPRequest)`: backwards incompat
- Supporting both the signature above and the current one: Seems overly complex
- Adding `path` to a second optional argument of a new `HTTPRequestOptions` interface to `request`: Best alternative, but also lacking much more substance--especially since `HTTPRequest` already supports additional meta options, like timeout.

In the future, we may want to reconsider redesigning these interfaces in a major version bump, but that is out of scope for this PR.

Definitely open the other implementation approaches!

These changes have dedicated unit tests. The unit tests validate the override logic as well as type enforcement.

Because these changes are backwards compatible, all existing code and tests should work as expected.

N/A

### Change types

### Checklist:

----

### DIFF
--- a/__tests__/unit/fetch-client.test.ts
+++ b/__tests__/unit/fetch-client.test.ts
@@ -11,7 +11,7 @@ import {
   QuerySuccess,
 } from "../../src";
 import { getDefaultHTTPClientOptions } from "../client";
-import { SupportedFaunaAPIs } from "../../src/http-client";
+import { SupportedFaunaAPIPaths } from "../../src/http-client";
 
 let fetchClient: FetchClient;
 
@@ -145,6 +145,21 @@ describe("fetch client", () => {
     }
   });
 
+  it("uses the default path if one is not provided in HttpRequest", async () => {
+    expect.assertions(1);
+    fetchMock.mockResponseOnce(JSON.stringify({}), {
+      headers: { "content-type": "application/json" },
+    });
+    await fetchClient.request({
+      ...dummyRequest,
+    });
+
+    expect(fetchMock).toHaveBeenCalledWith(
+      expect.stringContaining("/query/1"),
+      expect.any(Object),
+    );
+  });
+
   it("uses the path provided in the HttpRequest if provided", async () => {
     expect.assertions(1);
     fetchMock.mockResponseOnce(JSON.stringify({}), {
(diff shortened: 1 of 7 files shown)


__tests__/unit/fetch-client.test.ts
```diff
@@ -137,10 +138,25 @@ describe("fetch client", () => {
...
+    expect.assertions(1);
+    fetchMock.mockResponseOnce(JSON.stringify({}), {
+      headers: { "content-type": "application/json" },
+    });
+    await fetchClient.request({
+      ...dummyRequest,
+      path: "/non-the-default-api" as SupportedFaunaAPIs,
+    });
+    expect(fetchMock).toHaveBeenCalledWith(
+      expect.stringContaining("/non-the-default-api"),
```
echo-bravo-yahoo, line 158: the previous tests implicitly set the path to "/query/1", but none of them assert on it.

can you add an assertion so the default path case is tested?

echo-bravo-yahoo (dismissed):
i'm ok with this path forward (and with a major version bump that moves these to obj args as well).

how much blast radius is there in the docs site if we make a breaking change in these?

src/http-client/paths.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Readonly object representing the paths of the Fauna API to be used
+ * with HTTP clients.
+ */
+export const FaunaAPI = {
+  QUERY: "/query/1",
+  STREAM: "/stream/1",
+} as const;
+
+export type SupportedFaunaAPIs = (typeof FaunaAPI)[keyof typeof FaunaAPI];
```
ptpaterson, line 10: Nit question: should we name these `...Paths`? `FaunaAPIPaths` and `SupportedFaunaAPIPaths`?

ecooper:
> i'm ok with this path forward (and with a major version bump that moves these to obj args as well).
> 
> how much blast radius is there in the docs site if we make a breaking change in these?

Because these are largely internals they are really only covered by the docs that are auto-published here: https://fauna.github.io/fauna-js/2.2.0/

mwilde345 (approved):

PR #289, opened 2024-10-07
Add backwards compatible type params for HTTPRequest and HTTPStreamRequest

ecooper:
Data types for `HTTPRequest` and `HTTPStreamRequest` are hardcoded today. Because of #288, we need to support generics for `data`. They default to backwards compatible types.

### Description
This PR makes the following backwards compatible changes:

- `HTTPRequest` now takes a type parameter that is applied to `data`. It defaults to `QueryRequest`.
- `HTTPStreamRequest` now takes a type parameter that is applied to `data`. It defaults to `StreamRequest`.

### Motivation and context
To take advantage of `HTTPClient.request` against endpoints other than `/query/v1`, we need to support payloads other than `QueryRequest`.  While we don't have an immediate use case for extending `HTTPStreamRequest`, I would argue it makes sense to keep the approaches the same across our two primary request types (call/response and streaming).

No tests were added for this specific change. If all existing tests pass, this is a safe change. The underlying data types are never referred to or accessed directly within the client.

N/A

### Change types

### Checklist:

----

### DIFF
--- a/src/http-client/http-client.ts
+++ b/src/http-client/http-client.ts
@@ -6,15 +6,17 @@ import { SupportedFaunaAPIPaths } from "./paths";
 /**
  * An object representing an http request.
  * The {@link Client} provides this to the {@link HTTPClient} implementation.
+ *
+ * @typeParam T - The expected type of the request sent to fauna.
  */
-export type HTTPRequest = {
+export type HTTPRequest<T = QueryRequest> = {
   /**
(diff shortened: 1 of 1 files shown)


ptpaterson (approved):
Nice. I agree this will help keep things straight if we extend the client to reuse the "call/response" and "streaming" paths.

ecooper:
Closed in favor of #290.

PR #290, opened 2024-10-09
Add initial event feed support

ecooper:
This PR adds support for change feeds in the JS driver. To do this, we add a new `FeedClient` that can be initialized via `Client.feed(eventSource)`. `FeedClient` can then be used as a async generator for each page or event by event via `flatten`. `nextPage` is exposed for consumers who don't want to use the generator pattern.

```ts
const feed = client.changeFeed(fql`MyCollection.all().eventSource()`)

try {
  for await (const page of feed) {
    for (const event of page.events) {
      // ... handle event
    }
  }
} catch (error) {
  // ... handle fatal error
};
```

Or via flatten:

```ts
const feed = client.changeFeed(fql`MyCollection.all().eventSource()`)

for await (const user of feed.flatten()) {
  // do something with each event
}
```

### Description

Generally, this PR has surfaced the underlying abstractions aren't really equipped to support more Fauna endpoints. There's a refactoring opportunity across clients and the httpclient, but those changes are likely backwards incompatible. So I tried my best to follow the existing conventions in the codebase.

- `FeedClient`: Exposes change feed pagination via an async iterator via a public `nextPage` method. It also surfaces `flatten` to iterate through all events directly.
- `FeedClientConfiguration`: The required configuration for the iterator. Primarily, optional `start_ts` and `cursor`.
- `FeedPage`: Because there are errors that are embedded in `events` that are apparently fatal for the page, `FeedPage` wraps the events wire response in an iterator. If it sees an event type of "error", it throws. I'm not sure that's idiomatic behavior in JS (since you can continue to fetch pages after this event), but I kept parity with the python driver. Additionally, `FeedPage` provides public attributes to access meta data about the current page, such as the current cursor.
- `HTTPClient` generics: `HTTPClient` interfaces have been updated to support generics. This is the first new API using `request` other than query.
- `wire-protocol.ts`: Updated with the wire protocol as I understand it.
- `package.json`: Updated to use the change feed env flag in docker calls. This will need to be removed when the image is updated.

**Updates to naming**:

The following changes have been added to address the renaming of this feature:

* `EventSource`: An interface to represent the output of `eventSource()` and `eventsOn()`. `EventSource` replaces all publicly exported uses of `StreamToken`. `StreamToken` can be used anywhere `EventSource` is specified. This seemed the best way to move towards the correct language while keeping this feature backwards compatible.
* `eventSource()` and `eventsOn()`:  README and docstring uses of `toStream()` and `changesOn()` have been replaced with these new methods. Feed tests use these new FQL methods. Existing streaming tests have not been changes at all to prove backwards compatibility.

### Motivation and context

To support change feeds beta in the JS driver.

Unit, function, and integration tests were added for all new surface areas.

### Change types

### Checklist:

* - [x] My change requires a change to Fauna documentation.
* - [x] My change requires a change to the README, and I have updated it accordingly.

----


README.md
```diff
@@ -595,6 +598,129 @@ For supported properties, see
...
+### Change Feed options
+
+The client configuration sets the default options for `changeFeed()`. You can pass a ``ChangeFeedClientConfiguration` object to override these defaults:
+
+```ts
+const options: ChangeFeedClientConfiguration = {
+  long_type: "number",
+  max_attempts: 5,
+  max_backoff: 1000,
+  secret: "YOUR_FAUNA_SECRET",
```
jrodewig, line 716:
```suggestion
  secret: "FAUNA_SECRET",
```

README.md
```diff
@@ -595,6 +598,129 @@ For supported properties, see
...
+    }
+  }
+} catch (error: unknown) {
+  console.log("Non-retryable error:", error);
+}
+```
+
+### Change Feed options
+
+The client configuration sets the default options for `changeFeed()`. You can pass a ``ChangeFeedClientConfiguration` object to override these defaults:
```
jrodewig, line 709:
```suggestion
The client configuration sets the default options for `changeFeed()`. You can pass a `ChangeFeedClientConfiguration` object to override these defaults:
```

README.md
```diff
@@ -595,6 +598,129 @@ For supported properties, see
...
+```javascript
+const query = fql`Product.all().changesOn(.price, .stock)`;
+
+const changeFeed = client.changeFeed(query);
+```
+
+### Iterate on a Change Feed
+
+`changeFeed()` returns a `ChangeFeedClient` instance that can act as an `AsyncIterator`. You can
+use a `for await...of` to iterate through all the pages:
```
jrodewig, line 641:
```suggestion
use `for await...of` to iterate through all the pages:
```
Or:

```
a `for await...of` loop ...
```

jrodewig (commented):
Narrow +1 on the README aside from some nits. Thanks!

ptpaterson (changes requested):
This is looking great! I don't have any really major feedback. Closest question to major is how to handle `start_ts` and `cursor` input.

src/client.ts
```diff
@@ -852,6 +938,202 @@ export class StreamClient<T extends QueryValue = any> {
...
+    if (this.#clientConfiguration.page_size) {
+      req.data.page_size = this.#clientConfiguration.page_size;
+    }
+
+    // If we have a cursor, use that. Otherwise, use the start_ts if available.
+    if (this.#lastCursor) {
+      req.data.cursor = this.#lastCursor;
+    } else if (this.#clientConfiguration.start_ts) {
+      req.data.start_ts = this.#clientConfiguration.start_ts;
+    }
```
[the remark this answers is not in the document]
ecooper: If that's really the case, we can do some type changes to prevent it too. I'll do the type changes and the check and throw.

src/util/retryable.ts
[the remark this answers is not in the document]
ecooper: Yeah. Doing that in a follow-up. Didn't want to cause more of a blast radius than just change feeds.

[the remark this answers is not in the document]
ptpaterson: Yep!

src/client.ts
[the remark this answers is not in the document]
ecooper: jkjk. That makes the ergonomics of providing literals as configs a bit annoying. Added the type error as suggested!

ptpaterson (approved):
Let's goooo

src/client.ts
```diff
@@ -852,6 +938,209 @@ export class StreamClient<T extends QueryValue = any> {
...
+  }
+
+  async #resolveStreamToken(
+    fn: () => Promise<StreamToken>,
+  ): Promise<StreamToken> {
+    return await fn().then((maybeStreamToken) => {
+      if (!(maybeStreamToken instanceof StreamToken)) {
+        throw new ClientError(
+          `Error requesting a stream token. Expected a StreamToken as the query result, but received ${typeof maybeStreamToken}. Your query must return the result of '<Set>.toStream' or '<Set>.changesOn')\n` +
```
echo-bravo-yahoo, line 1101: `typeof` is going to be the run-time type, right? it may more useful to just safely stringify (`util.inspect`?) it instead; seems like it'll be "Object" in almost all misconfigurations.

src/util/retryable.ts
```diff
@@ -0,0 +1,30 @@
...
+export const withRetries = async <T>(
+  fn: () => Promise<T>,
+  {
+    maxAttempts,
+    maxBackoff,
+    shouldRetry = () => true,
+    attempt = 0,
+  }: RetryOptions,
+): Promise<T> => {
+  const backoffMs = Math.min(Math.random() * 2 ** attempt, maxBackoff) * 1_000;
```
echo-bravo-yahoo, line 17: this delays the _first_ request, right? at attempt 0, it's 2^0=1, so we're delaying by 0-1,000 ms.

should either special-case attempt 0 or pick a formula that evaluates to 0 naturally on attempt 0. maybe worth a unit test to validate/prevent regression.

echo-bravo-yahoo (changes requested):
~~marking as request changes to prevent merge - the only feedback i think is worth blocking over is retryable delaying request 0.~~

that's not an issue, removing the request changes flag.

__tests__/functional/change-feed-client-configuration.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  max_attempts: 3,
+  max_backoff: 20,
+  query_timeout_ms: 5000,
+  httpClient: defaultHttpClient,
+};
+const dummyStreamToken = new StreamToken("dummy");
+
+describe("ChangeFeedClientConfiguration", () => {
+  it("can be instantiated directly with a token", () => {
+    new ChangeFeedClient(dummyStreamToken, defaultConfig);
```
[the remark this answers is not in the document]
ptpaterson: I think not throwing is sufficient.

src/client-configuration.ts
```diff
@@ -204,6 +204,63 @@ export type StreamClientConfiguration = {
...
+   * The underlying {@link HTTPClient} that will execute the actual HTTP calls
+   */
+  httpClient: HTTPClient;
+
+  /**
+   * Controls what Javascript type to deserialize {@link https://docs.fauna.com/fauna/current/reference/fql_reference/types#long | Fauna longs} to.
+   *
+   * @see {@link ClientConfiguration.long_type}
+   */
+  long_type: "number" | "bigint";
```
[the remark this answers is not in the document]
ptpaterson: In beta, before we had the `long_type` config, we always deserialized Fauna `Long` types to BigInts, but we got feedback from folks that they want more control. We added a warning at the same time we added the new config.

https://github.com/fauna/fauna-js/blob/80b897eba674b965d5efc64a2897cf9c0db265f9/src/tagged-type.ts#L111-L120

Are you proposing that we effectively change that console.warn into thrown error? I think worth considering if we are going to make a major version bump at some point. But I also think it's outside the scope of this PR. I wouldn't want to make change/event feeds work differently than queries and streams.

src/client.ts
```diff
@@ -852,6 +938,209 @@ export class StreamClient<T extends QueryValue = any> {
...
+      throw new ClientError("The change feed has no more pages to fetch.");
+    }
+
+    const { httpClient } = this.#clientConfiguration;
+
+    const request = await this.#nextPageHttpRequest();
+    const response = await withRetries(() => httpClient.request(request), {
+      maxAttempts: this.#clientConfiguration.max_attempts,
+      maxBackoff: this.#clientConfiguration.max_backoff,
+      shouldRetry: (error) => error instanceof ThrottlingError,
```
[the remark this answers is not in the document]
ptpaterson: I don't think so. The only thing we use built-in retries for is 429 errors. Or at least, the consistent thing to do with the current query API is to retry only on 429 errors.

Looking beyond the scope of this PR, I still don't think we should, but probably worthwhile checking with the Core team if 503's are handled well enough to be considered retryable. You might think so, but historically, queries that result in 503's tend to do so consistently (i.e. not a good idea to retry).

src/client.ts
ptpaterson: Good point about `util.inspect`. Our domain types do not have toJSON overrides, so for example JSON.stringify(some_time_stub)` will result in something like `{ isoString: "2024-10-10T00:00:00Z" }`. So, at least `util.inspect` would show `TimeStub { isoString: "2024-10-10T00:00:00Z" }`.

How about using `${maybeStreamToken.constructor.name}` here and then `util.inspect` at the end of the message?

src/util/retryable.ts
ptpaterson: It tries to run `fn()` before any retries, so I think we are good there.

How _would_ we test? 🤔 

If we added an argument for a fn that calculates the delay, then we could test with a delay that doesn't use randomness and could depend on. Otherwise, I think we'd have to run a base test several times and check that the average execution time is below an amount.

I do think that customizable retries would be a great follow-up. Including how to pick which responses are retried and the algorithm for delays.

src/client-configuration.ts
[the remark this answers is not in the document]
echo-bravo-yahoo: i am proposing either:
- add a flag to upgrade from warn to throw (non-breaking) for all deserialization (change feeds, queries, and streams)
- changing the default behavior from warn to throw (breaking) for all deserialization (change feeds, queries, and streams)

this is likely going to be an insidious bug in business logic for customers at best, but because it affects non-stringified IDs, it can also bite in really simple, business-logic-agnostic use-cases like "a UDF that returns an ID, and a client that updates the thing at that ID."

src/client.ts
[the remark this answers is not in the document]
echo-bravo-yahoo: that's highly unusual - most web services retry on 500 because, while outages are often clustered and some requests will be poison pills, many 5xx are intermittent or only affect a subset of the fleet.

src/client.ts
echo-bravo-yahoo: i'm ok with that.

util.inspect has the added bonus that self-referential objects don't break the runtime.

src/util/retryable.ts
echo-bravo-yahoo: 👍 

Stubbing `math.random()` and `setTimeout` should work for testing a situation like this.

echo-bravo-yahoo (dismissed):

src/client-configuration.ts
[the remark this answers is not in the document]
ecooper: Let's solve that in a separate PR. I'll do the follow-up, and we can start with the backwards compat opt-in until the next major version.

src/util/retryable.ts
ecooper: I can add tests. NDB.

__tests__/functional/change-feed-client-configuration.test.ts
```diff
@@ -0,0 +1,80 @@
...
+  });
+
+  it("can be instantiated directly with a lambda", async () => {
+    new ChangeFeedClient(
+      () => Promise.resolve(dummyStreamToken),
+      defaultConfig,
+    );
+  });
+
+  it("throws a RangeError if 'max_backoff' is less than or equal to zero", async () => {
```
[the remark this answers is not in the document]
ecooper: Tested everything that's not protected by types.

__tests__/functional/change-feed-client.test.ts
```diff
@@ -0,0 +1,208 @@
...
+      ),
+    );
+
+    const changeFeed = new ChangeFeedClient(dummyStreamToken, defaultConfig);
+    await fromAsync(changeFeed.flatten());
+
+    expect(mockHttpClient.request).toHaveBeenCalledTimes(2);
+  });
+
+  it("throws an error for an error response", () => {
```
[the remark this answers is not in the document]
ecooper: Tested in retryable helper and in the integration tests.

src/client.ts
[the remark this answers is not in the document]
ecooper: Let's do a follow-up for this one.

src/client.ts
ecooper: I'll do this as a follow-up PR as it's going to change in two places (and one is unrelated to this PR). This is already a chonky PR.

__tests__/functional/change-feed-client.test.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    expect(mockHttpClient.request).toHaveBeenCalledWith(
+      expect.objectContaining({
+        client_timeout_ms: defaultConfig.query_timeout_ms,
+        headers: expect.objectContaining({
+          Authorization: "Bearer secret",
+          "x-format": "tagged",
+          "x-driver-env": expect.any(String),
+        }),
+        data: {
```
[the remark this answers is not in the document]
ecooper: Good catch. Tests prove it's there, too.

__tests__/functional/change-feed-client.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    expect(Array.from(secondPage.events)).toHaveLength(0);
+  });
+
+  it("uses a valid HTTPRequest", async () => {
+    const changeFeed = new ChangeFeedClient(dummyStreamToken, defaultConfig);
+    await fromAsync(changeFeed);
+
+    expect(mockHttpClient.request).toHaveBeenCalledWith(
+      expect.objectContaining({
+        client_timeout_ms: defaultConfig.query_timeout_ms,
```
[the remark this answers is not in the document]
ecooper: Done. I misunderstood the original timeout implementation for sure.

README.md
```diff
@@ -591,59 +591,59 @@ const options: StreamClientConfiguration = {
...
 in the API reference.
 
-## Change Feeds (beta)
+## Event Feeds
 
-The driver supports [Change Feeds](https://docs.fauna.com/fauna/current/learn/track-changes/streaming/#change-feeds).
+The driver supports [Event Feeds](https://docs.fauna.com/fauna/current/learn/cdc/#event-feeds).
 
-### Request a Change Feed
+### Request a Event Feed
```
echo-bravo-yahoo, line 605:
```suggestion
### Request an Event Feed
```

src/client-configuration.ts
```diff
@@ -205,47 +205,26 @@ export type StreamClientConfiguration = {
 };
 
 /**
- * Configuration for a change feed client. This typically comes from the `Client`
- * instance configuration.
+ * Configuration for a event feed client.
```
echo-bravo-yahoo, line 208:
```suggestion
 * Configuration for an event feed client.
```

src/client-configuration.ts
```diff
@@ -205,47 +205,26 @@ export type StreamClientConfiguration = {
...
+  Pick<
+    ClientConfiguration,
+    | "long_type"
+    | "max_attempts"
+    | "max_backoff"
+    | "client_timeout_buffer_ms"
+    | "query_timeout_ms"
+    | "secret"
+  >
+> & {
```
echo-bravo-yahoo, line 220: 👍 on type re-use.

is .cursor relevant? one of the examples in the README has cursor as optional on the FeedClientConfiguration: https://github.com/fauna/fauna-js/pull/290/files/9f19c10449e0c14c8d05e3e083292d4c8e1c3843..6b182a77a451c617c78bc00cfad246107267adc2#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R719

src/values/stream.ts
```diff
@@ -1,30 +1,42 @@
...
+
+export function isEventSource(value: any): value is EventSource {
+  if (typeof value.token === "string") {
+    return true;
+  }
+
+  return false;
+}
+
 export class StreamToken {
```
ptpaterson, line 40: Should `StreamToken` explicitly implement `EventSource` or is it sufficiently duck-typed?

ptpaterson: I mean... it IS duck typed. But I mean is that okay.

.github/workflows/pr_validate.yml
```diff
@@ -11,10 +11,14 @@ jobs:
         image: fauna/faunadb:latest
         ports:
           - 8443:8443
+        env:
+          FLAG_ACCOUNT_CHANGE_FEEDS: "true"
```
jrodewig, line 15: These flags aren't needed anymore. They're enabled by default on the Docker image.

README.md
```diff
@@ -34,6 +34,11 @@ See the [Fauna Documentation](https://docs.fauna.com/fauna/current/) for additio
     - [Iterate on a stream](#iterate-on-a-stream)
     - [Close a stream](#close-a-stream)
     - [Stream options](#stream-options)
+  - [Event Feeds](#event-feeds)
+    - [Request an Event Feed](#request-a-event-feed)
+    - [Iterate on an Event Feed](#iterate-on-a-event-feed)
+    - [Error handling](#error-handling)
+    - [Event Feed options](#event-feed-options)
```
jrodewig, line 41: Can we move this above the Event Streaming section? Bryan requested that we give Event feeds top billing.

src/client-configuration.ts
ptpaterson: `.cursor` is optional. If neither `.cursor` nor `.start_ts` are provided, then the feed will start effectively from MVT.

README.md
```diff
@@ -473,11 +476,11 @@ The driver supports [Event Streaming](https://docs.fauna.com/fauna/current/learn
...
 
 To get a stream token, append
-[`toStream()`](https://docs.fauna.com/fauna/current/reference/reference/schema_entities/set/tostream)
+[`eventSource()`](https://docs.fauna.com/fauna/current/reference/reference/schema_entities/set/eventsource)
 or
-[`changesOn()`](https://docs.fauna.com/fauna/current/reference/reference/schema_entities/set/changeson)
+[`eventsOn()`](https://docs.fauna.com/fauna/current/reference/reference/schema_entities/set/eventson)
 to a set from a [supported
-source](https://docs.fauna.com/fauna/current/reference/streaming_reference/#supported-sources).
+source](https://docs.fauna.com/fauna/current/reference/streaming_reference/#sets).
```
jrodewig, line 483: GH won't let me leave a suggestion, but can we change this to:

```
to a [supported Set](https://docs.fauna.com/fauna/current/reference/streaming_reference/#sets).

File: README.md [hunk 21]
```diff
@@ -473,11 +476,11 @@ The driver supports [Event Streaming](https://docs.fauna.com/fauna/current/learn
...
 
 To start and subscribe to the stream, pass the stream token to `stream()`:
```
jrodewig, line 485:
```suggestion
To start and subscribe to the stream, pass the event source to `stream()`:
```

README.md
```diff
@@ -473,11 +476,11 @@ The driver supports [Event Streaming](https://docs.fauna.com/fauna/current/learn
 ### Start a stream
 
 To get a stream token, append
```
jrodewig, line 478:
```suggestion
To get an event source, append
```

README.md
```diff
@@ -487,7 +490,7 @@ const response = await client.query(fql`
 
   {
     initialPage: set.pageSize(10),
-    streamToken: set.toStream()
+    streamToken: set.eventSource()
```
jrodewig, line 493:
```suggestion
    eventSource: set.eventSource()
```

We also need to update the below refs in this example.

README.md
```diff
@@ -498,7 +501,7 @@ client.stream(streamToken);
 You can also pass a query that produces a stream token directly to `stream()`:
```
jrodewig, line 501: GH won't let me leave a suggestion, but this should be:
```
You can also pass a query that produces an event source directly to `stream()`:
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
 in the API reference.
 
+## Event Feeds
+
+The driver supports [Event Feeds](https://docs.fauna.com/fauna/current/learn/cdc/#event-feeds).
+
+### Request a Event Feed
+
+An Event Feed asynchronously polls an [event stream](https://docs.fauna.com/fauna/current/learn/cdc/#event-feeds),
+represented by a stream token, for events.
```
jrodewig, line 608:
```suggestion
An Event Feed asynchronously polls an [event source](https://docs.fauna.com/fauna/current/learn/cdc/#create-an-event-source) for events.
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+
+To get a stream token, append `eventSource()` or `eventsOn()` to a set from a
+[supported source](https://docs.fauna.com/fauna/current/reference/streaming_reference/#supported-sources).
+
+To get paginated events for the stream, pass the stream token to
+`feed()`:
```
jrodewig, line 614:
```suggestion
To get paginated events, pass the event source to `feed()`:
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+  let set = Product.all()
+
+  {
+    initialPage: set.pageSize(10),
+    streamToken: set.eventSource()
+  }
+`);
+const { initialPage, streamToken } = response.data;
+
+const feed = client.feed(streamToken);
```
jrodewig, line 627:
```suggestion
    eventSource: set.eventSource()
  }
`);
const { initialPage, eventSource } = response.data;

const feed = client.feed(eventSource);
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+```
+
+You can also pass a query that produces a stream token directly to `feed()`:
```
jrodewig, line 630:
```suggestion
You can also pass a query that produces an event source directly to `feed()`:
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+
+```javascript
+const query = fql`Product.all().eventsOn(.price, .stock)`;
+
+const feed = client.feed(query);
+```
+
+### Iterate on a Event Feed
```
jrodewig, line 638:
```suggestion
### Iterate on an Event Feed
```

jrodewig (commented):
Thanks! Looks good overall. Few bigger pieces of feedback:

- We're using "event source" in place of "stream" and "stream token" now. I left comments for all the refs in the README, but we prolly also want to update this in the implementation.

- Bryan requested that we give Event Feeds top billing so I'd move that section before the Event Streaming one.

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+
+for await (const page of feed) {
+  console.log("Page stats", page.stats);
+
+  for (event in page.events) {
+    switch (event.type) {
+      case "update":
+      case "add":
+      case "remove":
+        console.log("Stream event:", event);
```
jrodewig, line 654:
```suggestion
        console.log("Event:", event);
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+```
+
+Alternatively, use `flatten()` to get paginated results as a single, flat array:
+
+```ts
+const query = fql`Product.all().eventsOn(.price, .stock)`;
+const feed = client.feed(query);
+
+for await (const event of feed.flatten()) {
+  console.log("Stream event:", event);
```
jrodewig, line 669:
```suggestion
  console.log("Event: ", event);
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+`);
+
+try {
+  for await (const page of feed) {
+    // Pages will stop at the first error encountered.
+    // Therefore, its safe to handle an event failures
+    // and then pull more pages.
+    try {
+      for (const event of page.events) {
+        console.log("Stream event:", event);
```
jrodewig, line 695:
```suggestion
        console.log("Event: ", event);
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+      }
+    } catch (error: unknown) {
+      console.log("Stream event error:", error);
```
jrodewig, line 698:
```suggestion
      console.log("Feed event error: ", error);
```

README.md
```diff
@@ -588,16 +591,140 @@ const options: StreamClientConfiguration = {
...
+    }
+  }
+} catch (error: unknown) {
+  console.log("Non-retryable error:", error);
```
jrodewig, line 702:
```suggestion
  console.log("Non-retryable error: ", error);
```

src/values/stream.ts
ecooper: I can add it explicitly. But interfaces match so TS is happy.

ecooper: When I fix all of the doc things I missed, I'll make this more explicit. I think duck typing has to be okay (and consider it a feature of interfaces), but we don't need to rely on it here.

README.md
```diff
@@ -562,7 +684,7 @@ const stream = await client.stream(fql`Product.all().eventSource()`);
 
 let count = 0;
 for await (const event of stream) {
-  console.log("Stream event:", event);
+  console.log("Event:", event);
```
jrodewig, line 687:
```suggestion
  console.log("Event: ", event);
```

README.md
```diff
@@ -539,7 +661,7 @@ stream.start(
       case "update":
       case "add":
       case "remove":
-        console.log("Stream event:", event);
+        console.log("Event:", event);
```
jrodewig, line 664:
```suggestion
        console.log("Event: ", event);
```

README.md
```diff
@@ -517,7 +639,7 @@ try {
...
```
jrodewig, line 642:
```suggestion
        console.log("Event: ", event);
```

README.md
```diff
@@ -469,36 +469,158 @@ const client = new Client({ http2_session_idle_ms: 6000 });
...
+```
+
+Alternatively, use `flatten()` to get paginated results as a single, flat array:
+
+```ts
+const query = fql`Product.all().eventsOn(.price, .stock)`;
+const feed = client.feed(query);
+
+for await (const event of feed.flatten()) {
+  console.log("Event:", event);
```
jrodewig, line 538:
```suggestion
  console.log("Event: ", event);
```

README.md
```diff
@@ -469,36 +469,158 @@ const client = new Client({ http2_session_idle_ms: 6000 });
...
+
+for await (const page of feed) {
+  console.log("Page stats", page.stats);
+
+  for (event in page.events) {
+    switch (event.type) {
+      case "update":
+      case "add":
+      case "remove":
+        console.log("Event:", event);
```
jrodewig, line 523:
```suggestion
        console.log("Event: ", event);
```

jrodewig (dismissed):
Narrow +1 on the docs. I left some nit suggestions, but they're not blocking IMO.

ptpaterson (approved):

PR #296, opened 2024-11-04
Validate cursor for feeds

ecooper:
This PR adds a validation check for cursor and makes sure it's a string at runtime. Non-string values can produce vague error messages otherwise.

### Description
This PR follows the current client configuration validation pattern to make sure `cursor` is a string when it is set.

### Motivation and context
While testing this change internally, we discovered if you managed to pass a non-string cursor during runtime, the error returned wasn't very helpful. It's probably easiest to mitigate this by validating it.

Added a test to the existing client configuration tests. The existing tests still pass.

### Change types

* - [X] Bug fix (non-breaking change that fixes an issue)

### Checklist:

* - [X] My code follows the code style of this project.

### DIFF
--- a/__tests__/functional/feed-client-configuration.test.ts
+++ b/__tests__/functional/feed-client-configuration.test.ts
@@ -97,4 +97,16 @@ describe("FeedClientConfiguration", () => {
       expect(e).toBeInstanceOf(RangeError);
     }
   });
+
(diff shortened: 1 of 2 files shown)


ptpaterson (approved):
🍬 LGTM

PR #298, opened 2024-11-04
Allow using a query and cursor together when calling client.feed()

ecooper:
This PR enables supplying a `cursor` value when using an unevaluated query with `client.feed()`.

### Description
This PR removes a check that prevented you from calling `client.feed()` with a `query` and a `cursor`. It is an unnecessary constraint, and event feeds will behave as you'd expect without this check.

### Motivation and context
This change allows you to resume at a given cursor while still passing in a query to generate the event source. Before this change, you'd need to first evaluate the query for the event source yourself before calling `feed` with `cursor`.

* An integration test on client was added
* Existing integration tests pass

### Change types

### Checklist:

----

### DIFF
--- a/__tests__/integration/feed.test.ts
+++ b/__tests__/integration/feed.test.ts
@@ -136,6 +136,46 @@ describe("Client", () => {
     expect(pages).toHaveLength(3);
     expect(Array.from(pages[0].events)).toHaveLength(1);
   });
+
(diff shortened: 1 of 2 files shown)


ptpaterson (approved):
I think this is a good thing to do. The new test covers the use case well.

jrodewig (approved):
👍 I'll update the README in a separate PR. Thanks!

PR #305, opened 2024-11-13
Better support for interfaces as query responses

### LINKED ISSUE #304: Simplify Typings
The current type system setup is unnecessarily complex and problematic. Given the expectation of an index signature and rather extraneous naming conventions, using TypeScript (TS) often results in a mild headache, requiring developers to meticulously code just to meet the driver's expectations. In most projects, `interface` is the preferred for type definition, with type aliases generally reserved for union or intersection types.

The driver's assumption of type alias expressions leads to cumbersome workarounds. Consider the following example:

```ts
interface Foo {
  a: string;
  b: boolean;
  c?: {
    d: string[]
  }
}
```

This interface is straightforward and should ideally integrate seamlessly with the driver's response model. However, the reality is less ideal:

```ts
const response = await client.query<Foo>(fql`...`);
```

<center>Error Screenshot</center>

This setup results in errors because the `QueryValue` constraint isn't met by our `Foo` interface. Even when we adjust as follows:

```ts
const response: QuerySuccess<Foo> = await client.query<Foo>(fql`...`);
```

The compiler still complains due to the constraints on `QueryValue`. The solution often involves extending the interface:

```ts
interface Foo extends Document {
  a: string;
  b: boolean;
  c?: {
    d: string[]
  }
}

const response: QuerySuccess<Foo> = await client.query<Foo>(fql`...`);
```

This works but introduces new issues:

1. **Assumption of Use**: It assumes `Foo` is exclusively for query usage, which might not always be the case.
2. **Type Narrowing**: Extending `Document` necessitates creating yet another type for accurate narrowing:

   ```ts
   interface FooResponse extends Foo, Document {}
   ```

This becomes problematic because:
- The `QuerySuccess` type exposes a `data` property, making index access types less viable without creating an additional interface.
- Using the original `Foo` interface elsewhere might lead to type conflicts due to the additional properties from `Document`.

---

I don't really understand why such exhaustive measures need to take place. What should be a simple generic type parameter has turned into:

1. Creating an additional interface to extend `Document`.
2. Modifying the function signature with the type parameter.
3. Using explicit type annotations.
4. Employing a generic type alias.

This process overly complicated and it's not necessary, furthermore what is the reasoning here? The generic can simply be augmented and resolve, there is no need for all this.

ptpaterson:
Provide better support and documentation for how to use interfaces as query responses.

resolves #304 

### Description
- Changes `QueryValueObject` from a `type` to an `interface`.
- Add some instruction for how to use the `QueryValueObject` interface to the README.

### Motivation and context
Sometimes, specifying interfaces for query responses is preferred over types. However, custom interfaces are currently not assignable to `QueryValue` and cannot be used where `QueryValue` is required (e.g. the `Client.query()` method).

See #304 

I added tests to demonstrate how interfaces may be used.

image

### Change types

### Checklist:

* - [x] My change requires a change to Fauna documentation.
* - [x] My change requires a change to the README, and I have updated it accordingly.

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ See the [Fauna Documentation](https://docs.fauna.com/fauna/current/) for additio
   - [Usage](#usage)
     - [Write FQL queries](#write-fql-queries)
     - [Typescript support](#typescript-support)
+      - [Interfaces as `QueryValue` responses](#interfaces-as-queryvalue-responses)
     - [Query options](#query-options)
     - [Query statistics](#query-statistics)
   - [Pagination](#pagination)
@@ -237,6 +238,25 @@ console.assert(userDoc.name === "Alice");
 console.assert(userDoc.email === "alice@site.example");
 
 client.close();
+```
+
+#### Interfaces as `QueryValue` responses
+
+To use a custom interface as a query response, extend the `QueryValueObject`
+interface.
+
+```typescript
+interface User extends QueryValueObject {
+  name: string;
+  email: string;
+}
+
+const query = fql`{
+  name: "Alice",
+  email: "alice@site.example",
+}`;
+
+const response: QuerySuccess<User> = await client.query<User>(query);
 ```
 
 ### Query options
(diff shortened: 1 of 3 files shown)


jrodewig (approved):
Narrow + 1 on README. Thanks!

cleve-fauna (approved):
Nice change. h/t @panoply

PR #308, opened 2024-11-21
util/logging.ts: Add support for logging

findgriffin:
util/logging.ts: Add LogHandler interface, LogLevel enum, and ConsoleLogHandler.

### Description

Add a log handler interface and implementation.

### Motivation and context

Following a LSE where we couldn't get reliable information from the customer side, we decided to add a logging facility to all of the drivers.

I spent a few hours surveying log library options in JavaScript/TypeScript. The conclusion that I came to is that they are either bulkier than what we need (log4js-node) or so simple that it's not worth us taking a dependency (pino), hence the strategy I adopted in this PR.

TODO: This PR does not yet actually call our log function from the request/response path, I thought the tests were failing on Monday. They appear to pass in GitHub Actions, so maybe that's a "fails on my machine" type of thing.

Unit tests?

### Change types

### Checklist:

----

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -626,6 +626,7 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
       this.#logger.debug(
         "Fauna HTTP %s Request to %s (timeout: %s), headers: %s",
         method,
+        this.#httpClient.getURL(),
         this.#clientConfiguration.endpoint.toString(),
         client_timeout_ms.toString(),
         JSON.stringify(headers),
@@ -641,7 +642,7 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
       this.#logger.debug(
         "Fauna HTTP Response %s from %s, headers: %s",
         response.status,
-        this.#clientConfiguration.endpoint.toString(),
+        this.#httpClient.getURL(),
         JSON.stringify(response.headers),
       );
 
@@ -1107,7 +1108,7 @@ export class FeedClient<T extends QueryValue = any> {
     this.#logger.debug(
       "Fauna HTTP %s Request to %s (timeout: %s), headers: %s",
       request.method,
-      this.#clientConfiguration.endpoint.toString(),
+      httpClient.getURL(),
       request.client_timeout_ms,
       JSON.stringify(request.headers),
     );
@@ -1119,7 +1120,7 @@ export class FeedClient<T extends QueryValue = any> {
     this.#logger.debug(
       "Fauna HTTP Response %s from %s, headers: %s",
       response.status,
-      this.#clientConfiguration.endpoint.toString(),
+      httpClient.getURL(),
       JSON.stringify(response.headers),
     );
 
--- a/src/client.ts
+++ b/src/client.ts
@@ -23,6 +23,7 @@ import {
   isHTTPResponse,
   isStreamClient,
   type HTTPClient,
+  HTTPResponse,
 } from "./http-client";
 import { Query } from "./query-builder";
 import { TaggedTypeFormat } from "./tagged-type";
@@ -103,6 +104,8 @@ export class Client {
   readonly #clientConfiguration: RequiredClientConfig;
   /** The underlying {@link HTTPClient} client. */
   readonly #httpClient: HTTPClient & Partial<HTTPStreamClient>;
+  /** A LogHandler instance. */
+  readonly #logger: LogHandler;
   /** The last transaction timestamp this client has seen */
   #lastTxnTs?: number;
   /** true if this client is closed false otherwise */
@@ -134,6 +137,12 @@ export class Client {
       endpoint: this.#getEndpoint(clientConfiguration),
       logger: this.#getLogger(clientConfiguration),
     };
+
+    if (clientConfiguration && clientConfiguration.logger) {
+      this.#logger = clientConfiguration.logger;
+    } else {
+      this.#logger = new ConsoleLogHandler(LOG_LEVELS.ERROR);
+    }
 
     this.#validateConfiguration();
 
@@ -362,6 +371,7 @@ export class Client {
       const streamClientConfig: StreamClientConfiguration = {
         ...this.#clientConfiguration,
         httpStreamClient: streamClient,
+        logger: this.#logger,
         ...options,
       };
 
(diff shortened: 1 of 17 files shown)


src/client.ts
```diff
@@ -556,6 +563,28 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+      typeof process === "object" &&
+      process.env &&
+      typeof process.env === "object"
+    ) {
+      if (faunaDebugEnabled(process.env["FAUNA_DEBUG"])) {
+        return new ConsoleLogHandler(LogLevel.DEBUG);
+      }
+    }
+    return new ConsoleLogHandler(LogLevel.ERROR);
+  }
```
ptpaterson, line 586: Anything that the user provides in the `logger` option is ignored, other than checking that it is undefined. We should return the explicitly given object if provided and fall back on the env variables

```suggestion
  #getLogger(partialClientConfig?: ClientConfiguration): LogHandler {
    if (
      partialClientConfig &&
      "logger" in partialClientConfig &&
      partialClientConfig.logger === undefined
    ) {
      throw new TypeError(`ClientConfiguration option logger must be defined.`);
    }
    
    if (partialClientConfig?.logger) {
      return logger
    }
    
    let env_debug: string | undefined = undefined;
    if (
      typeof process !== "undefined" &&
      process &&
      typeof process === "object" &&
      process.env &&
      typeof process.env === "object"
    ) {
      env_debug = process.env["FAUNA_DEBUG"]
    }
    
    if (faunaDebugEnabled(env_debug)) {
      return new ConsoleLogHandler(LogLevel.DEBUG);
    }
    
    return new ConsoleLogHandler(LogLevel.ERROR);
  }
```

src/util/logging.ts
```diff
@@ -0,0 +1,77 @@
...
+export interface LogHandler {
+  trace(msg?: string, args?: string[]): void;
+  debug(msg?: string, args?: string[]): void;
+  info(msg?: string, args?: string[]): void;
+  warn(msg?: string, args?: string[]): void;
+  error(msg?: string, args?: string[]): void;
+}
+
+export class ConsoleLogHandler implements LogHandler {
+  constructor(private readonly level: LogLevel) {
```
ptpaterson, line 43: prefer native private class properties over typescript magic ones. Declaring the field above the constructor will also ensure at a type level that it is assigned (which is currently not guaranteed).
```suggestion
  readonly #level: LogLevel;
  constructor(level: LogLevel) {
```

src/util/logging.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (level === null) {
+      this.level = LogLevel.ERROR;
+    }
```
ptpaterson, line 46: How should level be set if level is not provided or is null?

src/util/logging.ts
```diff
@@ -0,0 +1,77 @@
...
+ * @param fauna_debug - The String value of FAUNA_DEBUG.
+ */
+export function faunaDebugEnabled(fauna_debug: string | undefined): boolean {
+  if (fauna_debug === undefined || fauna_debug === null) {
+    return false;
+  } else {
+    if (fauna_debug.length == 0) {
+      return false;
+    } else {
+      if (fauna_debug.startsWith("0")) {
```
ptpaterson, line 25: If FAUNA_DEBUG is set to `"0"`, e.g. with `FAUNA_DEBUG=0`, then how does one configure a log level of TRACE?

ptpaterson (changes requested):
Logger looks good, but I don't think we can configure it correctly.

We should also have some unit tests

src/client.ts
ptpaterson: I _think_ we also want to pass in the log level provided to the `ConsoleLogHandler`, rather than hardcoding DEBUG and ERROR.

src/util/logging.ts
```diff
@@ -0,0 +1,77 @@
+export enum LogLevel {
```
ptpaterson, line 1: prefer a constant and companion union type over enum magic

ptpaterson:
We should put together some tests that demonstrate using logger stuff. But I've updated the things I have opinions and commented about.

src/client.ts
```diff
@@ -556,6 +563,33 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
...
+      process &&
+      typeof process === "object" &&
+      process.env &&
+      typeof process.env === "object"
+    ) {
+      const env_debug = parseDebugLevel(process.env["FAUNA_DEBUG"]);
+      return new ConsoleLogHandler(env_debug);
+    }
+
+    return new ConsoleLogHandler(LOG_LEVELS.ERROR);
```
ecooper, line 590: Are we saying we'd always be publishing console.log by default? That doesn't seem desirable.

ptpaterson: Good point. How about I add levels 5-CRITICAL and 6-NO_LOGGING and default to 6 instead of 4-ERROR.

This would align better with .NET, too https://github.com/fauna/fauna-dotnet/blob/f6142aa3adc431138c22b13b26cc6aba417c016b/Fauna/Util/Logger.cs#L55

ptpaterson: Added levels 5 and 6 and a `.fatal()` method to the logger interface. Default is now level 6.

src/client.ts
```diff
@@ -126,8 +135,15 @@ export class Client {
...
       secret: this.#getSecret(clientConfiguration),
       endpoint: this.#getEndpoint(clientConfiguration),
+      logger: this.#getLogger(clientConfiguration),
     };
 
+    if (clientConfiguration && clientConfiguration.logger) {
+      this.#logger = clientConfiguration.logger;
+    } else {
+      this.#logger = new ConsoleLogHandler(LOG_LEVELS.ERROR);
+    }
```
ecooper, line 145: `getLogger` returns a logger that's turned off by default (sounds good). But then we set a logger on the client itself at error (sounds bad)? Am I reading that right?

By default we want zero logging to the console.

ptpaterson: oh I missed that when adding the extra log levels. It ought to be OFF.

ptpaterson: Made logger a required configuration and updated calls to use that instead of keeping a copy in two places.

ecooper (approved):

PR #309, opened 2024-11-27
Add performance hints support

fauna-chase:
ENG-6989

### Description
This allows the driver to be configured to query with performance hints enabled. When enabled, performance hints will be returned in the summary of the response. 

### Motivation and context
Without this change we won't be able to display performance hints in the dashboard or CLI. Performance hints are intended to be a self service way for users to debug query performance/cost and just overall find ways to improve their queries.

I added a test for the specified header. I also tested this version of the CLI with a locally running version of the dashboard to ensure I could get performance hints to show up.

### Change types

*  [ x ] New feature (non-breaking change that adds functionality)

### Checklist:

* [ x ] My code follows the code style of this project.

----

### DIFF
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -121,6 +121,11 @@ export interface ClientConfiguration {
   typecheck?: boolean;
 
   /**
+   * Enable or disable performance hints. If no value is provided, performance hints will be disabled.
+   */
+  performanceHints?: boolean;
+
+  /**
    * Max attempts for retryable exceptions. Default is 3.
    */
   max_attempts?: number;
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -121,9 +121,9 @@ export interface ClientConfiguration {
   typecheck?: boolean;
 
   /**
-   * Enable or disable performance hints. If no value is provided, performance hints will be disabled.
-   * Performance hints will show up in the `summary` field of the QueryInfo object. This ends up as a top
-   * level field in the response object.
+   * Enable or disable performance hints. Defaults to disabled.
+   * The QueryInfo object includes performance hints in the `summary` field, which is a
+   * top-level field in the response object.
    */
   performance_hints?: boolean;
 
--- a/src/client-configuration.ts
+++ b/src/client-configuration.ts
@@ -122,8 +122,10 @@ export interface ClientConfiguration {
 
   /**
    * Enable or disable performance hints. If no value is provided, performance hints will be disabled.
+   * Performance hints will show up in the `summary` field of the QueryInfo object. This ends up as a top
+   * level field in the response object.
    */
-  performanceHints?: boolean;
+  performance_hints?: boolean;
 
   /**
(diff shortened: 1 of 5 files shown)


src/client-configuration.ts
```diff
@@ -120,6 +120,11 @@ export interface ClientConfiguration {
    */
   typecheck?: boolean;
 
+  /**
+   * Enable or disable performance hints. If no value is provided, performance hints will be disabled.
+   */
+  performanceHints?: boolean;
```
ecooper, line 126: As much as this pains me, for better or worse, these types have been using `snake_case` for configuration that gets passed as headers. We should keep these consistent and make this `performance_hints`.

We can move everything user-facing to camel in a future major version change. Until then, it's better to be consistent even if it's sort of silly.

src/client-configuration.ts
```diff
@@ -120,6 +120,11 @@ export interface ClientConfiguration {
    */
   typecheck?: boolean;
 
+  /**
+   * Enable or disable performance hints. If no value is provided, performance hints will be disabled.
```
ecooper, line 124: Can we add the context that this shows up in QueryInfo's summary? Otherwise, it's not very clear how this changes the output.

ecooper (approved):
Approved assuming you make the recommended changes.

src/client-configuration.ts
fauna-chase: o thats cool! I just didn't realize that and I just trend towards the camel case by default

src/client-configuration.ts
```diff
@@ -120,6 +120,13 @@ export interface ClientConfiguration {
...
+   * Performance hints will show up in the `summary` field of the QueryInfo object. This ends up as a top
+   * level field in the response object.
```
jrodewig, line 126: Just some minor nits. Take or leave as you see fit.
```suggestion
   * Enable or disable performance hints. Defaults to disabled.
   * The QueryInfo object includes performance hints in the `summary` field, which is a
   * top-level field in the response object.
```

src/wire-protocol.ts
```diff
@@ -96,6 +96,14 @@ export interface QueryOptions {
    */
   typecheck?: boolean;
 
+  /**
+   * Enable or disable performance hints. If no value is provided, performance hints will be disabled.
+   * Performance hints will show up in the `summary` field of the QueryInfo object. This ends up as a top
+   * level field in the response object.
```
jrodewig, line 102: Same changes as above:
```suggestion
   * Enable or disable performance hints. Defaults to disabled.
   * The QueryInfo object includes performance hints in the `summary` field, which is a
   * top-level field in the response object.
```

jrodewig (approved):
LGTM. I left some suggestions that you can take or leave as wanted. I'll ensure JS driver support is covered in https://github.com/fauna/docs.fauna.com/pull/2676.

ptpaterson (approved):
LGTM. Y'all did all the hard work!

PR #313, opened 2024-12-13
v10 driver - empty 200 response throws protocol error

mwilde345:
### Description

https://faunadb.atlassian.net/browse/FE-6237

Core router may return a 200 with an empty response, indicating an issue on their end. [See LSE](https://faunadb.atlassian.net/browse/LSE-205)
Don't try to parse that like we do a normal response, instead throw a protocol error.

Added a query test

Shell before:
Screenshot 2024-12-16 at 6 06 46 PM

Shell after:
Screenshot 2024-12-16 at 12 10 48 PM

### Change types

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -642,8 +642,8 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
       // Receiving a 200 with no body/content indicates an issue with core router
       if (
         response.status === 200 &&
-        response.body.length === 0 &&
-        response.headers["content-length"] === "0"
+        (response.body.length === 0 ||
+          response.headers["content-length"] === "0")
       ) {
         throw new ProtocolError({
           message:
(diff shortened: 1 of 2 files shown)


src/client.ts
```diff
@@ -639,6 +640,24 @@ in an environmental variable named FAUNA_SECRET or pass it to the Client\
         JSON.stringify(response.headers),
       );
 
+      // Receiving a 200 with no body/content indicates an issue with core router
+      if (
+        response.status === 200 &&
+        response.body.length === 0 &&
+        response.headers["content-length"] === "0"
+      ) {
+        throw new ServiceInternalError(
```
cleve-fauna, line 649: This is an error from Fauna's service; communcating that core issued a specific kind of error in the service.

ProtocolError would be more appropriate.

ptpaterson (approved):

PR #314, opened 2024-12-18
bump version

mwilde345:
Release changes from https://github.com/fauna/fauna-js/pull/313


ptpaterson:
This would also release logging changes, which I just realized breaks the HTTPClient interface. I think we can fix that though. I can do more in an about an hour to deal with that, but want to say something now.

ptpaterson (approved):

PR #323, opened 2025-02-05
Remove global import of 'DOM' type definitions. This is polluting customers environments by polluting the global type namespace.

cleve-fauna:
### Description
We are exporting the entire DOM lib as a library. Don't do that.
A recent commit added package-lock.json. We don't need that. Remove it.
### Motivation and context

This breaks customer environments as it pollutes the global type namespace.

I built with and without the changes here. You can see in the build the Triple-Slash-Directive used to be present but is no longer. Furthermore the linter I added throws in the original, but succeeds now.

Before:

Screenshot 2025-02-05 at 10 11 27 AM

After: 

Screenshot 2025-02-05 at 10 13 01 AM
### Change types

* - [x] Breaking change (backwards-incompatible fix or feature)

### Checklist:

----

### Out of Scope
The linter needs attention in general. It needs to move onto the latest version of eslint. Will work on that separately.


.eslintrc.js
```diff
@@ -13,9 +13,13 @@ module.exports = {
...
     sourceType: "module",
   },
-  plugins: ["@typescript-eslint", "eslint-plugin-tsdoc"],
+  plugins: ["@typescript-eslint", "eslint-plugin-tsdoc", "eslint-comments"],
   rules: {
     "@typescript-eslint/no-explicit-any": ["off"],
     "tsdoc/syntax": "error",
+    "@typescript-eslint/triple-slash-reference": [
+      "error",
+      { lib: "never", path: "never", types: "never" },
```
ptpaterson, line 22: good idea

ptpaterson (approved):
LGTM

PR #327, opened 2025-02-25
export `Query` class value so it can be `instanceof`'d

ssttevee:
### Description

Export the actual value of `Query` instead of just the type declaration so it can be used with `instanceof`.

### Motivation and context

I need to distinguish and handle `Query`s separately from other values.

To achieve this right now, I make an alias like this, which doesn't work well with typescript:
```
const Query = fql([]).constructor;
```

### Change types

### Checklist:

----


ptpaterson (approved):
Thanks @ssttevee! I agree marking `Query` as only a type was probably a mistake.

ssttevee:
@ptpaterson is there something blocking the workflows from running?

